PR #5, opened 2017-10-09
RAYMOND react hotel

ssozi91:
worked up to number 5 so far

### DIFF
--- a/src/components/StayedInDays.js
+++ b/src/components/StayedInDays.js
@@ -1,0 +1,10 @@
+import React from 'react';
+
+const  StayedInDays = (props) => {
+  var checkInDate=new Date(props.checkInDate);
+  var checkOutDate=new Date(props.checkOutDate);
+  var days = 1000*60*60*24;
+  var daysStayedIN = checkOutDate - checkInDate;
+  return Math.floor(daysStayedIN/days)
+}
+export default StayedInDays;
--- a/src/containers/Bookings.js
+++ b/src/containers/Bookings.js
@@ -7,7 +7,11 @@ import FakeBookings from "../data/fakeBookings.json";
 export default class Bookings extends Component {
   constructor(props) {
     super(props);
-    this.state = {isClicked:false};
+    this.state = {
+      isClicked: false,
+      bookings: FakeBookings,
+      selected  : false
+    };
   }
   search = () => {
     console.info("to do!");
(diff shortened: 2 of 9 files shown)


ssozi91:
got the days stayed row fixed

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
 import React, { Component } from "react";
 import Search from "../components/Search.js";
+import ResultsTable from '../components/ResultsTable'
```
kkarimi: Do you use Windows? though this might work on your machine this file is named `resultsTable.js` and you import it with capital here. Best to always use capital in react projects

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
...
+      return [];
+    }
+  }
+
+  onClicked = () => {
+    this.setState({
+     isClicked: true,
+   });
+  }
+  sortIt = (thing) => {
```
kkarimi: what is `thing`?! need to name things better

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
...
+    console.log('thing:', thing)
+    const bookings = this.state.bookings
+    const sortedBookings = this.state.bookings.sort()
```
kkarimi: You need to filter by the particular header that is clicked, there is no single way to sort a data table

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
...
+      this.setState({
+        bookings: bookings.reverse()
+      })
+    } else {
+      this.setState({
+        bookings: sortedBookings
+      })
+    }
+  }
+  // selectedRows = (){
```
kkarimi: Not good to push commented out code, delete them before pushing

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      isClicked: false,
+      bookings: FakeBookings,
+      selected  : false
+    };
+  }
   search = () => {
     console.info("to do!");
```
kkarimi: Why is this still console info?

src/containers/Bookings.js
```diff
@@ -1,18 +1,62 @@
...
+
+  handleBookings(){
+    if(this.state.isClicked===true){
+      return this.state.bookings
+    }else {
+      return [];
+    }
+  }
+
+  onClicked = () => {
```
kkarimi: onClick is not a good name for a function, I assume this is the search?

src/components/StayedInDays.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
kkarimi: This is not really a component, just a function. So you don't really need to import React here

PR #6, opened 2017-10-09
1-8 done

nabilwardeh:


src/containers/Bookings.js
```diff
@@ -1,19 +1,49 @@
...
+
 
 export default class Bookings extends Component {
-  search = () => {
-    console.info("to do!");
-  };
+  constructor(props) {
+    super(props);
+    this.state = { searched: false };
+    this.bookingData = null;
```
kkarimi: Although as you have spotted you can do it this way, per the React pattern this data belongs to the state, i.e.
`
this.state = { 
  searched: false,
  bookingData: null
 };
`

src/components/Results.js
```diff
@@ -0,0 +1,89 @@
...
+      filter: {},
+      rowsSelected: new Set()
+    };
+  }
+
+  orderData = (e) => {
+    if (!this.props.bookingResults) return;
+    let elementId = e.target.id;
+
+    if (this.state.filter[elementId]) {
```
kkarimi: I don't believe this `if ()` is ever true as you don't seem to update the `filter` in the state anywhere else? can you double check?

src/components/Search.js
```diff
@@ -1,41 +1,115 @@
...
+        className="form-control"
+        placeholder={this.props.placeholder}
+        name={this.props.name}
+        onChange={this.props.handleInput}
+      />
+    );
+  }
+}
+
+class SearchButton extends Component {
```
kkarimi: Using multiple components in a file can be a nice way to condense the number of files, but in a bigger project I'd say one file per component is probably best

src/containers/Bookings.js
```diff
@@ -1,19 +1,49 @@
...
+    } else if (filter.customerName) {
+      this.bookingData = FakeBookings.filter(booking => booking.firstName.includes(filter.customerName));
+    } else {
+      this.bookingData = [...FakeBookings];
+    }
+    this.addNumberOfDays();
+    this.setState({searched: true});
+  };
+
+  addNumberOfDays = () => {
```
kkarimi: Good job in breaking down the functions, makes it easier to read and write tests for

src/containers/Bookings.js
```diff
@@ -1,19 +1,49 @@
...
+  };
+
   render() {
     return (
       <div className="App-content">
         <div className="container">
-          <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
+          <Search search={this.search} runSearch={this.handleSearchButtonClicks} />
+          <Results bookingResults={this.bookingData} />
```
kkarimi: Should be per the commented out line, i.e. 
`<Results results={this.state.results} /> `

nabilwardeh:
Finished 9-10

PR #7, opened 2017-10-11
exercise 1 to 6 done

AmirTGitHub:


src/components/Button.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+const Button =  function(props){
+    // console.log(props)
+    return <button onClick={props.onClick}className="btn btn-primary fn-submit-name">{props.text}</button>
```
kkarimi: I recommend using `Format Document` in VSCode as you can see missing space here can break the code

src/components/Button.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+const Button =  function(props){
```
kkarimi: Good job in making smaller components, though keep in mind it is more consistent to use ES6 `const Button = (props) => {...}` to get used to it in the long run

src/components/RenderFakeData.js
```diff
@@ -0,0 +1,43 @@
...
+        <td>{jsonData.checkOutDate}</td>
+        <td>{moment(jsonData.checkOutDate).diff(jsonData.checkInDate, 'days')}</td>
+      </tr> 
+
+    )
+  });
+
+}
+
+  // const TableRow = function(props) {
```
kkarimi: Should leave the commented out code from the final version, if you use Git it should be in the history if you ever need it back

src/components/SearchFunction.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import Button from './Button';
+import SearchState from './SearchState.js'
+
+class SearchFunction extends React.Component {
```
kkarimi: Although you are breaking the components down pretty well, I'd say using names like `function` and `state` for components are confusing. It is better to think of components as visual components, so for example SearchForm and SearchResults would be better.

src/components/RenderFakeData.js
```diff
@@ -0,0 +1,43 @@
...
+      <tr key={i}>
+        <td>{jsonData.id}</td>
+        <td>{jsonData.title}</td>
+        <td>{jsonData.firstName}</td>
+        <td>{jsonData.surname}</td>
+        <td>{jsonData.email}</td>
+        <td>{jsonData.roomId}</td>
+        <td>{jsonData.checkInDate}</td>
+        <td>{jsonData.checkOutDate}</td>
+        <td>{moment(jsonData.checkOutDate).diff(jsonData.checkInDate, 'days')}</td>
```
kkarimi: Good job in using Moment!

PR #14, opened 2017-10-22
Ali's implementation

40thieves:
Just wanted to try my hand at an implementation 😄

### DIFF
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -3,13 +3,14 @@ import moment from 'moment'
 
 const ResultCount = props => (
   <div>
-    <span>Results ({props.results.length} found)</span>
+    <span>Results ({props.results.length} found, {props.selected.length} selected)</span>
   </div>
 )
 
 const TableHead = props => (
   <thead>
     <tr>
+      <th>Select</th>
       <th onClick={() => props.onSort('title')}>Title</th>
       <th onClick={() => props.onSort('firstName')}>First name</th>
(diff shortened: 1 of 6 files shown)


src/components/Results.js
```diff
@@ -0,0 +1,117 @@
...
+    })
+
+    this.setState({
+      results: sorted,
+      [sortField]: direction
+    })
+  }
+
+  handleSelected = (selectedEmail) => {
+    let found = this.state.selected.find((s) => s === selectedEmail)
```
40thieves: I think this could be cleaned up with a `Set`

src/components/Results.js
```diff
@@ -0,0 +1,117 @@
...
+    let newSelected
+    if (found) {
+      newSelected = this.state.selected.filter((s) => s !== selectedEmail)
+    } else {
+      newSelected = this.state.selected.concat(selectedEmail)
```
40thieves: This isn't clean. Just use a `Set`

src/components/Results.js
```diff
@@ -0,0 +1,117 @@
...
+      firstName: 'desc',
+      surname: 'desc',
+      email: 'desc',
+      roomId: 'desc',
+      results: props.results
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({ results: nextProps.results })
```
40thieves: Haven't you heard about hooks?

PR #15, opened 2018-02-28
React Week-1 Homework

prabshah:
React Homework 1, 2 and 3. I am still working on sorting the result bit!


package.json
```diff
@@ -8,7 +8,7 @@
     "react-scripts": "1.0.14"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "set PORT=2000 && react-scripts start",
```
40thieves: You don't actually need the `set` here 🙂

src/components/Results.js
```diff
@@ -0,0 +1,65 @@
+import React, { Component } from 'react';
+
+class Results extends Component {
+    render() {
+        return (
+            <div>
+                <table className="table search-table">
+                    <thead>
+                        <tr>
+                            <th>Search Results</th>
```
40thieves: I think this would be better off outside the `<table>`, just in a `<div>` perhaps.

src/components/Results.js
```diff
@@ -0,0 +1,65 @@
...
+                            <th />
+                            <th></th>
+                            <th />
+                        </tr>
+                        <tr>
+                            <th>Title</th>
```
40thieves: This works, but can you see how we could improve the readability a bit by extracting some more components? Maybe a `TableHeader`? There's at least 1 more component that might be extracted here.

src/components/Search.js
```diff
@@ -28,9 +30,10 @@ const Search = props => (
                 />
               </td>
               <td>
-                <button className="btn btn-primary fn-submit-name">Search IDs</button>
-              </td>
+                {<Button name = "Search IDs"/>}
```
40thieves: The `{` and `}` are unnecessary here

src/components/Search.js
```diff
@@ -28,9 +30,10 @@ const Search = props => (
...
+              </td> 
               {/* Add search by name here */}
+              <SearchByName />
```
40thieves: Nice job! Do you think it would be better to also extract a `SearchByCustomerId` component?

src/containers/Bookings.js
```diff
@@ -14,6 +14,7 @@ export default class Bookings extends Component {
         <div className="container">
           <Search search={this.search} />
           {/* <Results results={this.state.results} /> */}
```
40thieves: Might as well remove the comment here 🙂

package.json
prabshah: Many thanks for your reviews Ali. I have some other project running on my computer on port 3000 so I did this to change the port for this project.

src/components/layout/Header.js
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react'
```
40thieves: 👍 on organising the components into sub-directories

src/components/results/Results.js
```diff
@@ -0,0 +1,21 @@
...
+
+class Results extends Component {
+	render() {
+		var bookedItems = this.props.FakeBookings;
+
+		return (
+			<div>
+				<div> Search Reasults </div>
+				<table className="table">
+					<ResultsHeader bookings={bookedItems} />
```
40thieves: Not a huge problem, but it looks like you're not using the `bookings` prop in `ResultsHeader` so I wouldn't pass it

src/components/search/searchComponents/Button.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from "react";
+
+class Button extends Component {
+    
+    render() {
+        return (
+            <button className="btn btn-primary fn-submit-name" onClick={this.sayHello}>{this.props.name}</button>
```
40thieves: I don't think it's being used anywhere but your `onClick` won't do anything because `this.sayHello` isn't defined

src/containers/Bookings.js
```diff
@@ -1,22 +1,51 @@
...
-        <div className="container">
-          <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
-          <Results />
-        </div>
-      </div>
-    );
+  hideSearch = () => {
+    var btn = window.document.getElementById('search');
+    btn.innerHTML = 'Show Results'
```
40thieves: I would definitely recommend to not do this. It will have unintended consequences when React re-renders the search button

40thieves: On second look, I think actually you're passing props to set the button text so this line is unnecessary. I would remove it

src/components/search/searchComponents/SearchByCustomerId.js
```diff
@@ -0,0 +1,62 @@
...
+											className="form-control"
+											placeholder="Customer name"
+											name="customerName"
+										/>
+									</td>
+									<td>
+										
+										{/* { <button className="btn btn-info" id="clickMe" onClick={this.props.hideResults}>{this.props.text}</button> } */}
+										<button className="btn btn-info" id="search" onClick={
+											this.props.text  === 'Show Results' ? this.props.search : this.props.hideSearch
```
40thieves: I get what you're trying to do and it's not necessarily wrong, however I would usually just use 1 click handler function that is passed down from `Bookings` instead of switching based on the `text` prop.

Then the click handler function in `Bookings` would check the current state and change whether we're showing the results or not. Something like:

```js
search = () => {
  if (this.state.text === 'Show Results') {
    // ...
  } else {
    // ...
  }
}
```

src/containers/Bookings.js
```diff
@@ -1,22 +1,51 @@
...
+	render() {
+    var props = {
+      search : this.search,
+      hideSearch : this.hideSearch,
+      text : this.state.text
+    }
+		return (
+			<div className="App-content">
+				<div className="container">
+					<Search {...props} />
```
40thieves: 👍 for using `...props` for spreading!

src/components/results/resultsComponents/ResultsBody.js
```diff
@@ -1,15 +1,49 @@
...
+	// 	if(index > -1) {
+	// 		selectedItems.splice(index, 1)
+	// 	}else {
+	// 		selectedItems.push(key)
+	// 	}
+	// 	this.setState({
+	// 		selectedItems
+	// 	})
+	// }
+	getRowColor = (key) => {
```
40thieves: I like this idea! I think it makes table rows much more readable 👍 

However, you can actually do this in CSS and it's a bit better for performance. Because we're using bootstrap you can just add the `.table-striped` class to the table. Here's the relevant bit in the documentation: https://getbootstrap.com/docs/3.3/css/#tables-striped. Note the bit that mentions the `nth-child` selector - this is how bootstrap does this in CSS

PR #16, opened 2018-03-01
firsThreeFromHomeWork

ameerkabir:


src/components/Search.js
```diff
@@ -7,35 +7,10 @@ const Search = props => (
...
-                <input
-                  id="customerId"
-                  type="text"
-                  className="form-control"
-                  placeholder="Customer id"
-                  name="customerId"
-                />
-              </td>
-              <td>
-                <button className="btn btn-primary fn-submit-name">Search IDs</button>
```
dahfool: @ameerkabir You need to only move the button out not the whole. 
Change the button on the `Search.js` page to be a separate component

src/components/SortableHeaders.js
```diff
@@ -0,0 +1,103 @@
+import React, { Component } from "react";
+
+class SortableHeaders extends Component {
```
40thieves: I wouldn't have thought of doing it this way, but I really like this 👍

src/containers/Bookings.js
```diff
@@ -1,61 +1,75 @@
...
-  //   console.info("searchName",event.target.id);
-  // }
-  // setSearchId = event =>{
-  //   console.info("searchId",event.target.id);
-  // }
-  // searchAction = event =>{
-  //   console.info(event.target.id);
-  // }
+  componentDidMount() {
+    fetch("http://localhost:3030/api/customers-and-reservations")
```
40thieves: Nice work!

PR #17, opened 2018-03-01
i have done the first 3 parts of homework

monaazami:


src/components/Button.js
```diff
@@ -0,0 +1,15 @@
+import React ,{Component} from 'react';                 
+
+
+   export default class Button extends Component{
+       render(){
+           return(
```
dahfool: @monaazami well done, you have done completed the tasks.

src/components/Search.js
```diff
@@ -28,14 +30,58 @@ const Search = props => (
...
+              <th>Search by name</th>
+              <th />
+              <th></th>
+              <th />
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>
+                <input
```
dahfool: The input fields for searchs are very similar, its not part of the exercise but could be move into a separate component as well.

src/containers/Bookings.js
```diff
@@ -4,16 +4,33 @@ import Results from "../components/Results.js";
...
+  searchId = () => {
+    this.setState({results:FakeBookings}) 
     console.info("to do!");
   };
+  searchName = () => {
+    this.setState({results:FakeBookings}) 
+    console.info("not to do!");
+  };
+  sortBy = (colum)=>{
+    FakeBookings.sort((a,b)=>{if(a[colum]>b[colum]){return 1}})
```
dahfool: @monaazami To debug why `number of days` is not filtering properly, put a console.log(colum) in the sortBy function and check what is the value of `colum` when you press `number of days`

src/containers/Bookings.js
```diff
@@ -4,33 +4,79 @@ import Results from "../components/Results.js";
...
+        }
+      }
+      ),
+      direction: this.state.direction === 'asc'
+        ? 'desc'
+        : 'asc'
+    })
   }
-  
+  toggleSelected = (x) => {
```
40thieves: Having the selected state up in bookings might be a bit problem later - because you would toggle selected state for all items in the list, not just the one you selected. This state might be better further down the tree

src/containers/Bookings.js
```diff
@@ -4,33 +4,79 @@ import Results from "../components/Results.js";
...
-      this.state= {results:[]};
-
+    this.state = { results: [], direction: 'asc', style: {} };
 
+    this.addNumberOfDays(FakeBookings);
   }
+
+
+  addNumberOfDays = (FakeBookings) => {
+    FakeBookings.map((item) => {
```
40thieves: I think this would work, but it's not quite *semantic*. If we want to loop through all the list items and add something to them (like we are here), then I would use a `.forEach`. We use `.map` if we're creating a **new** array that is based on the old array, but we're transforming all of the items in the list.

PR #18, opened 2018-03-03
Eugene's PR

yevhensydorov:


src/components/Search.js
```diff
@@ -1,4 +1,6 @@
 import React from 'react';
+import { SearchButton } from './SearchButton';
+import { CustomersTable } from './custTable.js';
```
40thieves: The convention is to name component files exactly the same as the component name, so in this case `CustomerTable`. And the `.js` is unnecessary here - it'll assume you meant `.js` if you leave it off

src/components/Search.js
```diff
@@ -28,14 +30,30 @@ const Search = props => (
                 />
               </td>
               <td>
-                <button className="btn btn-primary fn-submit-name">Search IDs</button>
+                <SearchButton />
+              </td>
+              <td>
```
40thieves: Do you see a pattern here? Would it be useful to extract out some components to make it more readable?

src/components/Search.js
40thieves: Also, generally the convention is to use `export default` on your components. So then here you would import like `import SearchButton from ...`

src/components/SearchButton.js
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react';
+
+export class SearchButton extends Component {
```
40thieves: As mentioned above, generally you want to `export default` components

src/components/custTable.js
```diff
@@ -0,0 +1,77 @@
...
+            <th>First Name</th>
+            <th>Surname</th>
+            <th>Email</th>
+            <th>Room id</th>
+            <th>Check in date</th>
+            <th>Check out date</th>
+          </tr>
+        </thead>
+        <tbody id="rows">
+          <tr>
```
40thieves: There is a another pattern here 🙂 Would extracting a component be useful? This might also help with upcoming homework 😉

src/components/Customers.js
```diff
@@ -0,0 +1,61 @@
...
+	render() {
+		// console.log(this.state.columnToSort, this.state.sortDirection);
+		return (
+			<div>
+				<p>Find Customers: 
+					<button className='btn btn-primary' onClick={this.getData}>Find</button>
+				</p>
+				<table className="table">
+					<TableHead sort={this.handleSort} />
+					<TableData customers={orderBy(
```
40thieves: Hmm 😄 Lodash is handy here, but can you implement your own sort method? Hint: there is `Array.sort`!

src/App.css
```diff
@@ -50,4 +50,143 @@ tr {
...
+
+.default {
+  background-color: #fff;
+}
+
+.hidden {
+  display: none;
+}
+
+/*SPINNER STYLES*/
```
40thieves: Nice 😂

src/components/Customers.js
```diff
@@ -88,24 +90,72 @@ class Customers extends React.Component {
...
+			return (
+				<p>There is an error: <span>{this.state.error}</span></p>
+			);
+		} else {
+			return (
+				<div>
+					<section className="get-results">
+
+						{/*SPINNER SECTION*/}
+	          <section className={this.state.isLoading ? 'spinner' : 'hidden'}>
```
40thieves: Maybe this would be worth extracting out?

And I would usually just remove the spinner from the DOM (i.e. not render it) if we're not loading. It's simpler to do this than hide with CSS

PR #19, opened 2018-03-03
React homework 1-3

OlenaKashuba:


src/components/CustomersTable.js
```diff
@@ -0,0 +1,39 @@
...
+			            		<th>First name</th>
+			            		<th>Surname</th>
+			            		<th>Email</th>
+			            		<th>Room id</th>
+			            		<th>Check in date</th>
+			            		<th>Check out date</th>
+			            	</tr>
+			         </thead>
+			         <tbody>
+			          		<tr>
```
40thieves: Can you see an opportunity to make things a bit more readable here? Could we extract a component? Looking ahead a bit to next week's homework, it might be useful to have some way of rendering lots of rows using some data 😄

src/components/Search.js
```diff
@@ -28,14 +30,26 @@ const Search = props => (
                 />
               </td>
               <td>
-                <button className="btn btn-primary fn-submit-name">Search IDs</button>
+                <SearchButton />
+              </td>
+              <td>
```
40thieves: There's a repeating pattern here, maybe something we want to make into a component or 2?

src/components/Results.js
```diff
@@ -0,0 +1,62 @@
...
+class Results extends React.Component {
+	constructor(props) {
+		super(props)
+
+		this.state = {
+			data: [],
+			isFetched: false
+		}
+
+		this.checkIfFetched = this.checkIfFetched.bind(this);
```
40thieves: Because we are using create-react-app, we can use the class properties syntax and avoid the `.bind` call here (it's a bit ugly 🙂). So it would look like:

```js
checkIfFetched = (event) => {
  ...
}
```

src/components/Results.js
```diff
@@ -0,0 +1,62 @@
...
+		checkIfFetched(event) {
+			console.log('checkIfFetched is here');
+			if (this.state.isFetched === false) {
+				this.getData();
+			}
+		}
+		
+
+		getData() {
+			console.log('this is', FakeBookings);
```
40thieves: As far as I can tell, there's no reason to split the `checkIfFetched` and `getData` methods?

src/components/SearchInputHeader.js
```diff
@@ -0,0 +1,11 @@
+ import React from 'react';
+
+class SearchInputHeader extends React.Component {
+	render () {
+		return (
+			<th>{this.props.header}</th>
```
40thieves: In all honesty, I wouldn't make a component if it's this small. But it's not necessarily a problem.

I think it would benefit from being made into a functional component

src/components/TableBody.js
```diff
@@ -0,0 +1,32 @@
...
+class TableBody extends React.Component {
+	constructor(props){
+		super(props);
+
+	}
+	render() {
+		const data = this.props.fetched;
+
+		let rowWithData = data.map(function(item) {
+			return <tr>
```
40thieves: Just to be on the safe side, I wrap multi line returns like this in parens `(` and `)`

src/components/TableBody.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+class TableBody extends React.Component {
+	constructor(props){
+		super(props);
+
+	}
+	render() {
+		const data = this.props.fetched;
```
40thieves: I'm not a huge fan of the `fetched` prop name here. I would go with something like `bookings`. Or if you're trying to make `TableBody` more generic maybe `items` would be better.

src/components/Results.js
```diff
@@ -0,0 +1,62 @@
...
+import TableHead from './TableHead';
+import TableBody from './TableBody';
+import * as FakeBookings from "../data/fakeBookings.json";
+
+
+class Results extends React.Component {
+	constructor(props) {
+		super(props)
+
+		this.state = {
```
40thieves: Using state in `Results` works for now - because we're just listening for a click in the results table. But in the future when we start trying to do a real search with the input in the `Search` component, I think you'll find that pulling the state up to the `Bookings` component is better.

40thieves:
Sorry I forgot - now try out exercises 6 & 7. I'll give a hint that you might want to look at moment.js for exercise 6. The `.diff` method will help you compare dates 🙂

src/components/SearchInputHeader.js
OlenaKashuba: i think thats a problem of table layout. if I did something similar from scratch, i wouldn't make it a separate component, just in this case I somehow decided it will look nicer

40thieves: 😆 yep, one of the things on my todo list is to change the example so that it doesn't use `<table>` for layout

src/components/Results.js
OlenaKashuba: I separated them to make clear that handlers do different things - just check and fetch. But you are right, they can be in one

src/components/TableBody.js
OlenaKashuba: done!

src/components/TableBody.js
OlenaKashuba: agree, done

src/components/Results.js
OlenaKashuba: not sure what you mean, sorry

src/components/Results.js
OlenaKashuba: that's not so obvious for me now

src/components/Results.js
40thieves: So your code here is valid. But having to add `this.whatever = this.whatever.bind(this)` for every method in your code is a bit of a pain.

So there is a new way to do the same thing in JavaScript. This new syntax is enabled because the homework was created with create-react-app.

If you remove the `.bind` line above, then change the `checkIfFetched` method to use the (slightly different) syntax that I pasted above, it's a bit nicer to read and you don't have extra lines. It's minor but if you have lots of methods in a component it can make it easier to read.

src/components/Results.js
40thieves: Ok, try changing your code so that when you click on one of the search buttons, it loads your results. This is more like what we actually would want to make in a real application - show no results until the user searches for something.

You'll probably find that it's quite difficult if you keep the state in `Results`.

OlenaKashuba: yeah, i got it, i have state in results because the button i click (i created it) belongs to results block. once we use input fields, which belong to another sibling component, i will want to set state to a parent block

src/components/TableBody.js
```diff
@@ -18,6 +19,7 @@ class TableBody extends React.Component {
 						<td> {item.roomId} </td>
 						<td> {item.checkInDate} </td>
 						<td> {item.checkOutDate} </td>
+						<td> {moment(item.checkOutDate).diff(item.checkInDate, 'days')}</td>
```
40thieves: 👍

PR #20, opened 2018-03-04
Louis' s CYF-HOTEL_REACT

LouiaR:


src/components/Search/Search.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import SearchButton from './SearchButton.js';
+import TablTitle from './SearchHeading.js';
```
40thieves: This is a bit confusing to me - if you're importing a component, I would expect the variable to be named the same as the component name

src/components/Search/SearchHeading.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const tablTitle = (props) => {
```
40thieves: The convention is to use exactly the same name for the component here as the name of the file. So it would be either `SearchHeading`, or rename the file to `TablTitle`. Also remember that the convention is to name components with a capital letter to start.

Personally I would pick `SearchHeading`. I think it's more descriptive, and I like explicit names rather than shorted names like "tabl"

src/components/Search/Search.js
```diff
@@ -0,0 +1,32 @@
...
+      <table className="table search-table">
+        <thead>
+          <tr>
+            <TablTitle title='Customer id' /><th />
+            <TablTitle title='Customer Name' /><th />
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <SearchInput id='customerId' placeholder='Customer id' name='customerid' />
```
40thieves: If you have lots of props, I like to format my code more like this:

```js
<SearchInput
  id='customerId'
  placeholder='Customer id'
  name='customerid'
/>
```

src/containers/Bookings.js
```diff
@@ -1,19 +1,79 @@
 import React, { Component } from "react";
-import Search from "../components/Search.js";
-// import Results from "../components/Results.js";
-// import * as FakeBookings from "../data/fakeBookings.json";
+import Search from "../components/Search/Search.js";
```
40thieves: I like that you've organised your components into sub-directories!

src/containers/Bookings.js
```diff
@@ -1,19 +1,79 @@
...
+import Results from "../components/Results/Results.js";
+import FakeBookings from '../data/fakeBookings.json'
+
 
 export default class Bookings extends Component {
+  state = {
```
40thieves: This looks like next week's homework! I'm going to leave off reviewing it for now, maybe you can check it once we've gone over props & state in the class.

src/index.css
```diff
@@ -3,3 +3,11 @@ body {
   padding: 0;
   font-family: sans-serif;
 }
+
+.hidden{
```
40thieves: I can't find any usages of the `.hidden` class? If it's not used, it might be better to remove it.

src/index.css
```diff
@@ -3,3 +3,11 @@ body {
...
+  display: none;
+}
+
+.fn-submit-name{
+  cursor: pointer;
```
40thieves: It looks like this class is being applied to elements which have the `.btn` class from Bootstrap. These should already get `cursor: pointer` styles, so this is unnecessary.

src/components/Search/Search.js
LouiaR: Yeah, you're right, I rename it

src/components/Search/SearchHeading.js
LouiaR: Ditto. Done

src/components/Search/Search.js
LouiaR: Done

src/containers/Bookings.js
LouiaR: Thank you.

src/index.css
LouiaR: Done

src/index.css
LouiaR: It doesn't show the pointer when I press the button that's why I added it.

LouiaR: Thank you for your comments

dahfool: @LouiaR good work on the last commit, tidying the code, better indentation, remove unnecessary comments and renaming for clarity.

src/containers/Bookings.js
```diff
@@ -27,52 +30,71 @@ export default class Bookings extends Component {
...
     this.setState({
       sort: !isSorted
     });
-    this.sortColumn(val);
+    this.sortColumn();
+  }
+  
+  selectRowHandler = (e) => {
+    {e.currentTarget.className !== 'row-hover'? 
+      ( e.currentTarget.className = this.state.background,
```
40thieves: I think this is mutating the DOM directly. This is something we want to avoid in React as any changes we make can be accidentally wiped by re-rendering. We should use state and tell React how to render the DOM instead.

PR #21, opened 2018-03-04
CYF-HOTEL-REACT_Ibrahim

ibramoha:


src/components/Search.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-
+import Btn_id from "./Btn_id.js";
```
40thieves: The naming convention for components is to use `CamelCasing` not `Snake_casing`

src/components/SearchResult.js
```diff
@@ -0,0 +1,71 @@
...
+              <th>Email</th>
+              <th />
+              <th>Room id</th>
+              <th>Check in date</th>
+              <th>Check out date</th>
+            
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
```
40thieves: Can you see an opportunity for extracting a component in the future here? Is there a repeating pattern?

ibramoha:
Thank you ali.

src/components/SearchResult.js
```diff
@@ -33,6 +33,7 @@ const SearchResult = props => (
                 <td>{row.roomId}</td>
                 <td>{row.checkInDate}</td>
                 <td>{row.checkOutDate}</td>
+                <td >{parseInt((new Date(row.checkOutDate) - new Date(row.checkInDate)) / (1000 * 60 * 60 * 24))} days</td>
```
40thieves: I actually quite like this approach, it's pretty readable. And good job on splitting up the time into readable pieces 👍

However you might want to look into the moment.js library. It has a `.diff` method which is helpful for this sort of thing

src/containers/Bookings.js
```diff
@@ -1,70 +1,108 @@
...
+    let doesShow = this.state.show;
+    this.setState({
+      data: this.state.newData.rows,
+      show: !doesShow
+    });
+  };
+
+  colourToggleHandler = e => {
+    if (e.currentTarget.className === "Highlite") {
+      e.currentTarget.className = this.state.bd;
```
40thieves: Hmm, this is manipulating the DOM directly. We try to avoid doing this in React because if we re-render after we've manipulated, React will not realise it's changed and remove our changes. We should use state to keep track of whether a row is selected or not.

src/containers/Bookings.js
```diff
@@ -1,70 +1,108 @@
...
 
-
-  toggleHandler = () => {
-    console.log(FakeBookings);
-
-    let doesShow = this.state.show;
-    this.setState( { data: FakeBookings,
-                     show: !doesShow } );
+  componentDidMount() {
+    fetch(`http://localhost:8080/api/reactclass`)
```
40thieves: Nice job! 🎈

src/containers/Bookings.js
ibramoha: yaaaaaaa Alright! I will use state.

PR #22, opened 2018-03-04
Mohsen cyf-hotel-react

Moradi66:


src/containers/Bookings.js
```diff
@@ -1,21 +1,64 @@
...
+    this.state = {
+      FakeBookings: '',
+      term: '',
+      result: false,
+      direction: {
+        sortin: 'asc'
+      },
+      text: ''
+    }
+    this.searchHandler = this.searchHandler.bind(this)
```
40thieves: You can avoid having to do this (somewhat ugly) `.bind` by changing your methods to use the class properties syntax:

```js
searchHandler = (inputType) => {
  ...
}
```

src/containers/Bookings.js
```diff
@@ -1,21 +1,64 @@
...
+          ? (a[key] < b[key])
+          : (b[key] < a[key])
+      )),
+      direction: {
+        [key]: this.state.direction[key] === 'asc'
+          ? 'desc'
+          : 'asc'
+      }
+    })
+    this.searchHandler(this.state.text)
```
40thieves: Is this call necessary? Sorting shouldn't affect the other state?

src/database/hotel.sql
```diff
@@ -0,0 +1,117 @@
+create table invoices (
```
40thieves: Why are you including a SQL file? We don't have any way of accessing the data?

src/components/results/ResultBody.js
```diff
@@ -0,0 +1,21 @@
...
+            {filteredResult.map((data, i) =>
+                <tr key={i}>
+                    <td>{data.title}</td>
+                    <td>{data.firstName}</td>
+                    <td>{data.surname}</td>
+                    <td>{data.email}</td>
+                    <td>{data.roomId}</td>
+                    <td>{data.checkInDate}</td>
+                    <td>{data.checkOutDate}</td>
+                    <td>{parseInt((new Date(data.checkOutDate) - new Date(data.checkInDate)) / (86400000), 10)} days</td>
```
40thieves: Oh I forgot to say - I guess this works, I wouldn't say it's very readable. I would recommend looking at installing moment.js and using the `.diff` method

src/containers/Bookings.js
```diff
@@ -2,22 +2,37 @@
...
     this.state = {
-      FakeBookings: '',
-      term: '',
+      FakeBookings: [],
+      term: null,
       result: false,
       direction: {
-        sortin: 'asc'
+        sortin: 'asc',
+        error: null,
```
40thieves: I wouldn't put the `error` state within `direction` state. Those aren't really related so I'd split them apart

src/containers/Bookings.js
```diff
@@ -2,22 +2,37 @@
...
     this.searchHandler = this.searchHandler.bind(this)
     this.sortBy = this.sortBy.bind(this)
+    this.sortByNumber = this.sortByNumber.bind(this)
+  }
+
+
+  componentDidMount() {
+    fetch(`http://192.168.43.95:8080/api/Customersjoinreservation`)
+      .then((res) => res.json())
+      .then((FakeBookings) => { this.setState({ FakeBookings: FakeBookings.customers }) })
```
40thieves: The bookings aren't fake now 🙂. So I would rename this to `bookings` or maybe `data`.

src/containers/Bookings.js
```diff
@@ -26,28 +41,46 @@ export default class Bookings extends Component {
...
+      direction: {
+        [key]: this.state.direction[key] === 'asc'
+          ? 'desc'
+          : 'asc'
+      },
+      text: this.state.text
+    })
+  }
+
+  sortByNumber(key) {
```
40thieves: I quite like your solution 👍 It's very explicit, which is great. There is an alternative where you just have 1 `sort` method and it checks whether the field that you're sorting by is a number or not. This would reduce some of the duplicated code that you have here.

But I'm not saying that is a "more correct" answer. There's a tradeoff here between the explicit-ness of your solution and the more complicated but less duplicated solution that I'm suggesting. You might not appreciate this but this is great!! We're having the kind of conversation that me & my colleagues have 👍

src/containers/Bookings.js
```diff
@@ -2,22 +2,37 @@
...
+      FakeBookings: [],
+      term: null,
       result: false,
       direction: {
-        sortin: 'asc'
+        sortin: 'asc',
+        error: null,
       },
-      text: ''
+      text: null,
```
40thieves: I think you're missing a loading state? When you start your HTTP fetch, you're not showing anything to the user. So they might be confused about what is happening (and might click the button lots of times)

PR #27, opened 2018-05-12
Islam's homework

islam95:


islam95:
Could I get a review for this homework please?

src/containers/Bookings.js
```diff
@@ -1,19 +1,62 @@
...
+import BookingsMessage from "../components/BookingsMessage.js";
+import Results from "../components/Results.js";
+import fakeBookings from "../data/fakeBookings.json";
+import Counter from "../components/Counter";
 
 export default class Bookings extends Component {
-  search = () => {
-    console.info('TO DO!');
+  state = {
+    count: 0,
```
kabaros: I am not sure what count1, count2 are? can they have better names to describe what they do? same for `input`.

src/components/Button.js
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from 'react';
+
+class Button extends Component {
+    render(){
+        return <button onClick={this.props.bookingsById} className="btn btn-primary fn-submit-name">{this.props.name}</button>;
```
kabaros: This is subjective. But I don't see the point of separating Button into its own component, it should just be inside Search component. It is not a generic button, and it only does one thing related to the search functionality.

src/components/Results.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import moment from "moment";
+
+const Results = props => {
+  const fakeBookings = props.results.map(booking => {
+    let checkInDate = moment(booking.checkInDate);
```
kabaros: nice using `moment`

kabaros: it can be `const`, no need for `let`

src/containers/Bookings.js
```diff
@@ -1,19 +1,62 @@
...
+  };
+
   render() {
     return (
       <div className="App-content">
         <div className="container">
-          <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
+          <BookingsMessage />
+          <Search userInput={this.userInput} bookingsById={this.bookingsById} />
```
kabaros: same for `userInput` and `bookingsById` - name them more descriptive. Something like `onKeywordChange` and `onSearch` (and I think both of them can just be one function onSearch maybe? you can just get the value of the text box when they click the button if you're not filtering the list on the fly)

src/containers/Bookings.js
```diff
@@ -1,19 +1,62 @@
...
+
+  bookingsById = () => {
+    this.setState({
+      bookings: fakeBookings.filter(booking => {
+        return booking.id === this.state.input;
+      })
+    });
+  };
+
+  increase = () => {
```
kabaros: what you've done here is correct in previous versions of React but there is a new recommended way in the current version. https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous (setState can take a function that has the previous state passed to it as an argument)

src/containers/Bookings.js
```diff
@@ -1,19 +1,62 @@
...
-  
+
+  userInput = e => {
+    e.preventDefault();
+    this.setState({
+      input: parseInt(e.target.value, 10)
+    });
+  };
+
+  bookingsById = () => {
```
kabaros: naming - I would call it `getBookingsById` or `filterBookings` (have a verb in the beginning of the name) since it's doing something (setting the state). But that is my opinion. Just be consistent, and name things the same way.

src/containers/Bookings.js
```diff
@@ -1,19 +1,62 @@
...
-          <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
+          <BookingsMessage />
+          <Search userInput={this.userInput} bookingsById={this.bookingsById} />
+          Results: {this.state.bookings.length}
+          <Results results={this.state.bookings} />
+          Sum: {this.state.sum}
+          <br />
+          {this.state.count}
+          <Counter name="Counter 1" increase={this.increase} />
```
kabaros: again .. I would name the prop `onIncrease` to make it obvious to people that it is an event.

kabaros:
Also add `propTypes` while you're at it :)

src/components/Button.js
islam95: I don't know, it was on the task to create a separate component for a button. I think it's just to show us how to create new component, because at that time we had no idea about React components.

src/containers/Bookings.js
islam95: If I join them into one function I am not sure if I can take value from an input after pressing the button. What I am doing here is that I am passing these two function to Search component and in Seach component there are input and button to which I pass each function. I don't remember if I can know the value of input in onClick function of a button.

src/containers/Bookings.js
islam95: Would it be like this: 
```
this.setState((prevState, props) => ({
  count: ++prevState.count,
  sum: ++prevState.sum
}));
```
And do I need to use props argument here?

src/containers/Bookings.js
islam95: Good suggestions, thanks!

src/containers/Bookings.js
kabaros: yeah it should look this

src/containers/Bookings.js
kabaros: you're right .. forget what I said about getting value. Just name them with something more clear and related to Search

src/containers/Bookings.js
islam95: When you try to choose a meaningful variable name: 
https://instagram.com/p/BjcOSE8jo67/

src/containers/Bookings.js
islam95: This is just rubbish, it was just for myself. I just left them there for a future reference.

PR #32, opened 2018-06-12
Lesson 1.3

vikrep:


robota-x:
Good PR, i really like how you subclassed the components, very clean. The search functionality is not really working and it's not making use of the state, but that's something we can discuss in person as it would be difficult to address in a PR.

Few things you can act on:
* The PR errors locally when you try to run it after a git clone. This is because you added a dependency, `moment.js`. While you (I suppose) installed it locally, it was not saved to the package.json file, so when I tried to install the requirements on my machine, moment.js was missing. Find out how to save requirements for your node projects
* Once that is fixed, the project will still not run, due to a error in Counter, TableData and all the elements making use of result. Can you find out why and fix that?

PR #35, opened 2018-09-09
Add solutions to homework one

rajgthub:
I attach my solutions to homework-exercise one here with this pull request. Please have a look at the answers and give your feedback when you get time. Thanks  very much for  spending your time on checking my work! I have not focused much on the styling bit.... I should have separated forms into component! I also used state to add a little bit of interaction to the UI. If you want me to do without states, I can do it again. Let me know your thoughts!

### DIFF
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -1,6 +1,7 @@
 import React from "react";
 import "../Results.css";
-import { TableHeader, TableRow } from "./table";
+import TableHeader from "./TableHeader"
+import TableRow from "./TableRow"
 const Results = ( {results} ) => {
     // console.log('testing...', results)
   if (results && results.length > 0) {
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -2,7 +2,7 @@ import React from "react";
 import "../Results.css";
 import { TableHeader, TableRow } from "./table";
 const Results = ( {results} ) => {
-    console.log('testing...', results)
+    // console.log('testing...', results)
   if (results && results.length > 0) {
     return (
       <table className="Results">
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -1,0 +1,18 @@
+import React from "react";
+import "../Results.css";
+import { TableHeader, TableRow } from "./table";
+const Results = ( {results} ) => {
+    console.log('testing...', results)
+  if (results && results.length > 0) {
+    return (
+      <table className="Results">
+        <TableHeader />
+        {results.map((result, key) => <TableRow key= {key} {...result} />)}
+      </table>
+    );
+  } else {
+    return <h1>No results found!</h1>;
+  }
+};
+
+export default Results;
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -1,40 +1,27 @@
-import React from 'react';
-import Buttton from './Button'
-const Search = ({search}) => (
+import React from "react";
+import SearchControls from "./SearchControls";
+const searchById = {
+  htmlFor: "customerId",
+  searchBy: "customer Id",
+  id: "customerId",
+  placeholder: "Customer Id",
+  btn_label: "Search IDs"
+};
+const searchByName = {
+  htmlFor: "customerName",
+  searchBy: "customer Name",
+  id: "customerName",
+  placeholder: "Customer Name",
+  btn_label: "Search Names"
+};
+const Search = ({ searchHandler }) => (
   <div className="search">
     <div className="page-header">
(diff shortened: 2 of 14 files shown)


src/components/table.js
```diff
@@ -0,0 +1,48 @@
...
+      <td>{surname}</td>
+      <td>{email}</td>
+      <td>{roomId}</td>
+      <td>{checkInDate}</td>
+      <td>{checkOutDate}</td>
+      <td>{dateDifference(checkInDate, checkOutDate)}</td>
+    </tr>
+  </tbody>
+);
+export { TableHeader, TableRow };
```
40thieves: Nice! In this case I agree with your decision to keep the components in the same file as they are related to each other. But the convention generally is to put components in separate files.

src/containers/Bookings.js
```diff
@@ -1,19 +1,23 @@
...
-
+import React, { Component } from "react";
+import Search from "../components/Search.js";
+import Results from "../components/Results.js";
+import rsults from "../data/fakeBookings.json";
 export default class Bookings extends Component {
+  state = { rsults: undefined };
   search = () => {
-    console.info('TO DO!');
+    this.setState({ rsults});
```
40thieves: 👍 We're going to flesh out this function more in upcoming exercises once we've looked at state in class.

src/components/table.js
```diff
@@ -0,0 +1,48 @@
...
+    <tr>
+      <td>{id}</td>
+      <td>{title}</td>
+      <td>{firstName}</td>
+      <td>{surname}</td>
+      <td>{email}</td>
+      <td>{roomId}</td>
+      <td>{checkInDate}</td>
+      <td>{checkOutDate}</td>
+      <td>{dateDifference(checkInDate, checkOutDate)}</td>
```
40thieves: Nice use of a function to clean up here!

src/components/Results.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "../Results.css";
+import { TableHeader, TableRow } from "./table";
+const Results = ( {results} ) => {
+    console.log('testing...', results)
+  if (results && results.length > 0) {
+    return (
+      <table className="Results">
+        <TableHeader />
+        {results.map((result, key) => <TableRow key= {key} {...result} />)}
```
40thieves: Good that you included a `key` 👏

src/components/Search.js
```diff
@@ -10,8 +10,28 @@ const Search = props => (
...
+              id="customerId"
+              className="form-control"
+              placeholder="Customer Id"
+            />
+            <Buttton label="Search IDs" search= {search}/>
+          </div>
+        </div>
+      </div>
+      {/* search by name*/}
+      <div className="col">
```
40thieves: This isn't part of the exercises, but do you notice there's some duplication here - what could you do to improve that?

rajgthub:
Thanks very much for the feedback and the time you spent on checking my solutions even if you were so busy with your work! I will reply to your comments when I get home.

src/components/table.js
rajgthub: Ok

src/components/Search.js
rajgthub: You are right! I have done the changes as you pointed out. I hope I did not make it more complicated!

rajgthub:
Hi Ali,

Thanks very much for checking my solutions. Yes, I will delete the table.js
as I am not using it anymore.
Best wishes,
Raju

src/containers/Bookings.js
```diff
@@ -1,13 +1,36 @@
...
+          return res;
+        } else {
+          throw new Error("HTTP error");
+        }
+      })
+      .then(res => res.json())
+      .then(({ results }) => {
+        console.log("kehtnn", results);
+        bookings = results;
+        this.setState({ isLoading: false, filteredRsults: results });
```
40thieves: One quick tip - it's probably a good idea to clear out any errors you have in state when you receive a successful response. That way you don't accidentally show an error message even if there is a successful response

src/containers/Bookings.js
```diff
@@ -57,27 +80,35 @@ export default class Bookings extends Component {
...
+        ) : (
+          !this.state.error && "No results found!!"
+        )}
+      </div>
+    );
+    const templateLoading = <h1>successfully completed!</h1>
+    /* I am using two different identifier: 
+    database has names like first_name and API returns as firstName; 
+    not enogh time! 
+    I will change it later*/
```
40thieves: 👍 It looks like you're on the right track though!

PR #36, opened 2018-09-12
solution for first react homework CYF hotel react BY Hasanein Amori

hasaninqh:


src/components/Results.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import FakeBookings from '../data/fakeBookings.json';
```
robgmerrill: You are off to a great start! However, Exercise 4 states: Instead of hard-coding data in the Bookings component, load data from the fakeBookings.json file. Hint: look in the Bookings component for how to import data from a JSON file. **Pass it to the Results component using props** and show all of the fake bookings.

So for this challenge you will want to import `import FakeBookings from '../data/fakeBookings.json';` into the `Bookings.js` component.

You will then want to pass it to the `Results` component via props. 

In your `Bookings` component you would have something like `<Result booking={FakeBookings} />`. This will take in the imported `FakeBookings` json data and pass it as a prop that you can access in your `Results` component within `booking` that you can then map over as you have done. 

So the different in this file is that you want line 10 to have `const Result = (props) => {` then line 28 would be something like `{props.booking.map(item => (`

Let me know if you have any questions!

src/containers/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import Search from '../components/Search.js';
-// import Results from '../components/Results.js';
-// import FakeBookings from '../data/fakeBookings.json';
```
robgmerrill: Make sure to bring this line of code back in as I mentioned in the the `Results.js` file.

src/containers/Bookings.js
```diff
@@ -13,7 +13,7 @@ export default class Bookings extends Component {
       <div className="App-content">
         <div className="container">
           <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
+          <Results />
```
robgmerrill: As I mentioned in the `Results.js` file we want to pass in the imported fake bookings json with props to the `Results` component to meet the exercise 4 challenge requirements. You can do this with something like `<Results booking={FakeBookings} />

src/components/Results.js
hasaninqh: Thanks so much, it was very informative feedback. Very appreciated

PR #37, opened 2018-09-12
homework lesson one

Behnazz:
Just finished Lesson one homework.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -38,11 +38,20 @@
 .search {
   padding: 5px 0 20px 0;
 }
+tr{
+  border-top: 1px solid black;
 
-tr {
-  color: #5b5757;
 }
-
+td{
+  color: #5b5757;
+  padding-left: 25px;
+  text-align: center
+}
+th{
+  padding-left: 25px;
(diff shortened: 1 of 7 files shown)


src/App.css
```diff
@@ -38,11 +38,20 @@
 .search {
   padding: 5px 0 20px 0;
 }
+tr{
+  border-top: 1px solid black;
 
-tr {
+}
+td{
```
robgmerrill: Looking good. This is a very small detail but for your `tr`, `th`, `td` selectors you should probably have a space between the selector and the opening curly bracket :)

src/containers/Bookings.js
```diff
@@ -1,19 +1,19 @@
...
   };
-  
+
   render() {
     return (
       <div className="App-content">
         <div className="container">
           <Search search={this.search} />
-          {/* <Results results={this.state.results} /> */}
+          <Result booking={FakeBookings} />
```
robgmerrill: Great job on passing in the in the json file in the `booking` prop for the `Result` component :)

PR #38, opened 2018-09-13
all homework week 20

daleel1:


src/components/Results.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import moment from "moment";
+
+const Results = (props) => {
+  const res = props.results;
```
Mullefa: Not sure if this intermediate value is needed? Could just do `props.results.map()`

src/components/Results.js
```diff
@@ -0,0 +1,37 @@
...
+          <th>title</th>
+          <th>First Name</th>
+          <th>Surname </th>
+          <th>Email </th>
+          <th>Room ID </th>
+          <th>Check in Date </th>
+          <th>Check out Date </th>
+          <th>number of days</th>
+        </tr>
+        {res.map(function (result){
```
Mullefa: Think an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) is more idiomatic.

src/components/Results.js
```diff
@@ -0,0 +1,37 @@
...
+        return (<tr>
+          <td>{result.title}</td>
+          <td>{result.firstName}</td>
+          <td>{result.surname}</td>
+          <td>{result.email}</td>
+          <td>{result.roomId}</td>
+          <td>{result.checkInDate}</td>
+          <td>{result.checkOutDate}</td>
+          <td>{moment(result.checkOutDate).diff(moment(result.checkInDate), 'days')}</td>
```
Mullefa: Nice! Maybe cleaner to factor it out as a utility function though e.g. `dayDiff()`

src/components/Search.js
```diff
@@ -11,9 +12,18 @@ const Search = props => (
           <label htmlFor="customerId">Customer Id</label>
           <div className="search-row">
             <input type="text" id="customerId" className="form-control" placeholder="Customer Id" />
-            <button className="btn btn-primary fn-submit-name">Search IDs</button>
+            <Button name="Search IDs"/>
```
Mullefa: `buttonText` might be a clearer / more accurate prop key name than `name`

src/containers/Bookings.js
```diff
@@ -1,21 +1,24 @@
 import React, { Component } from 'react';
 import Search from '../components/Search.js';
 // import Results from '../components/Results.js';
-// import FakeBookings from '../data/fakeBookings.json';
+import Results from '../components/Results.js';
 
+// import FakeBookings from '../data/fakeBookings.json';
```
Mullefa: Comment (`// ...`) can be removed

src/components/Results.js
Mullefa: A better function argument name could be `reservation` or `booking` (rather than `result`)

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "moment": "^2.22.2",
```
Mullefa: Nice!

reber12: @Mullefa thanks for your comments on my PR, and for taking the time to review my work

PR #39, opened 2018-09-14
hotel-react

than2042:
Hi Ali, 
Thank you. Here is my Homework. Any advice welcome! 

TT

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,0 +1,9 @@
+import React from 'react';
+
+const ButtonSearch = (props) => (
+ <button className="btn btn-primary fn-submit-name">
+     {props.name} {props.id}
+    </button>
+)
+
+export default ButtonSearch;   
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import moment from 'moment';
+//import FakeBookings from '../data/fakeBookings.json';
+
+const ResultTable = props => (
+  <table>
+    <thead>
+      <tr>
+        <th>id</th>
+        <th>title</th>
+        <th>First Name</th>
+        <th>Surname</th>
+        <th>email</th>
+        <th>roomId</th>
+        <th>Check In Date</th>
+        <th>Check Out Date</th>
+        <th>Number Of Stay</th>
+      </tr>
+    </thead>
+    <tbody>
+      {
+        props.data.map(row => (
+          <tr>
+            <td>{row.id}</td>
+            <td>{row.title}</td>
+            <td>{row.firstName}</td>
+            <td>{row.surname}</td>
+            <td>{row.email}</td>
+            <td>{row.roomId}</td>
+            <td>{row.checkInDate}</td>
+            <td>{row.checkOutDate}</td>
+            <td>{diffDays(row.checkInDate, row.checkOutDate)}</td>
+          </tr>
+        ))
+      }
+
+    </tbody>
+  </table>
+)
+
+function diffDays (checkInDate, checkOutDate) {
+ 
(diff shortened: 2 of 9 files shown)


src/components/Search.js
```diff
@@ -11,12 +12,31 @@ const Search = props => (
           <label htmlFor="customerId">Customer Id</label>
           <div className="search-row">
             <input type="text" id="customerId" className="form-control" placeholder="Customer Id" />
-            <button className="btn btn-primary fn-submit-name">Search IDs</button>
+            <ButtonSearch id="Search Ids" />
```
40thieves: Nice use of props to pass down the search id!

src/components/Results.js
```diff
@@ -0,0 +1,51 @@
...
+        <th>email</th>
+        <th>roomId</th>
+        <th>Check In Date</th>
+        <th>Check Out Date</th>
+        <th>Number Of Stay</th>
+      </tr>
+    </thead>
+    <tbody>
+      {
+        props.data.map(row => (
```
40thieves: Looks good!

src/components/Results.js
```diff
@@ -0,0 +1,51 @@
...
+          <tr>
+            <td>{row.id}</td>
+            <td>{row.title}</td>
+            <td>{row.firstName}</td>
+            <td>{row.surname}</td>
+            <td>{row.email}</td>
+            <td>{row.roomId}</td>
+            <td>{row.checkInDate}</td>
+            <td>{row.checkOutDate}</td>
+            <td>{diffDays(row.checkInDate, row.checkOutDate)}</td>
```
40thieves: I like the function to clean up the component 👍

src/components/Button.js
```diff
@@ -1,9 +1,46 @@
...
+class Button extends React.Component {
+  constructor() {
+    super()
+    this.state = {
+      value: ""
+    }
+  }
+
+  handleUpdate = (e) => {
+    this.setState({ value: e.target.value })
```
40thieves: Because you're listening for updates on the `<input>` and updating the state, this pattern that you're using is called a "controlled component". There's not really a problem with using this pattern, but it is slightly unusual for this particular case. It is often used if we need to transform or keep track of exactly what the user has typed.

React has an alternative pattern (called "uncontrolled component"), were we don't add a `onChange` handler, and then just get the input when we click the button.

I've just realised that we haven't looked a feature of React that will make this easier in class yet (called *refs*). Hopefully that will make more sense on how we can do that after the class.

src/components/Button.js
```diff
@@ -1,9 +1,46 @@
 import React from 'react';
 
-const ButtonSearch = (props) => (
- <button className="btn btn-primary fn-submit-name">
-     {props.name} {props.id}
-    </button>
-)
+class Button extends React.Component {
```
40thieves: It looks like the purpose of this component has changed - it no longer only renders a button. So if I was trying to use `<Button />` then I would be a bit confused. I think the best solution here would be to rename this component to reflect the job that it's actually doing. Can you think of a good component name?

src/containers/Bookings.js
```diff
@@ -2,30 +2,68 @@ import React, { Component } from 'react';
...
+            filterName: " ",
+            FakeBookings: [],
+            filteredBookings: []
+        }
+    }
+
+    componentWillMount() {
+        this.setState({
+            FakeBookings,
+            filteredBookings: FakeBookings
```
40thieves: Interesting use of `componentWillMount`. We're going to cover this in detail in the class. In this case, personally I would initialise the state in the constructor instead. 

However, in the next homework, we will need to use `componentWillMount` like similar to this.

PR #40, opened 2018-09-14
 Homework completed for review

Seraphineyoung:
Homework for React week 1.

Kindly review and give feedbacks.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,0 +1,10 @@
+import React from "react";
+
+const ButtonSearch = props => {
+  return (
+    <button onClick={props.search} className="btn btn-primary fn-submit-name">
+      {props.buttonText}
+    </button>
+  );
+};
+export default ButtonSearch;
--- a/src/components/Results.js
+++ b/src/components/Results.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { Component } from "react";
 import moment from "moment";
 
 const diffDays = (checkInDate, checkOutDate) => {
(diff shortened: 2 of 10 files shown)


src/components/Button.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+const ButtonSearch = props => {
+  return (
+    <button onClick={props.search} className="btn btn-primary fn-submit-name">
```
40thieves: Good work implementing the click handler 👍

src/components/Search.js
```diff
@@ -1,22 +1,48 @@
...
+              <ButtonSearch search={props.search} buttonText="Search Id" />
+            </div>
           </div>
         </div>
       </div>
-    </div>
-  </div>
-);
 
+      <div className="row search-wrapper">
```
40thieves: This is not part of the homework, but there is some duplication here. How could we improve this?

src/containers/Bookings.js
```diff
@@ -1,19 +1,23 @@
...
+import Results from "../components/Results.js";
+import FakeBookings from "../data/fakeBookings.json";
 
 export default class Bookings extends Component {
+  state = { reservations: [] };
+
   search = () => {
-    console.info('TO DO!');
+    console.info("to do");
+    this.setState({ reservations: FakeBookings });
```
40thieves: Well done implementing state to show the bookings! We'll cover state in more detail in the class

src/components/Results.js
```diff
@@ -0,0 +1,50 @@
...
+          <tr>
+            <td>{reservation.title}</td>
+            <td>{reservation.firstName}</td>
+            <td>{reservation.surname}</td>
+            <td>{reservation.email}</td>
+            <td>{reservation.checkInDate}</td>
+            <td>{reservation.checkOutDate}</td>
+            <td>{reservation.roomId}</td>
+            <td>
+              {diffDays(reservation.checkInDate, reservation.checkOutDate)}
```
40thieves: I like the use of a function to clean up the day diff!

src/components/Button.js
Seraphineyoung: Thanks Ali

src/containers/Bookings.js
```diff
@@ -4,19 +4,39 @@ import Results from "../components/Results.js";
...
+    const myResult = FakeBookings.filter(singleReservation => {
+      console.log(UserInput);
+      console.log(singleReservation.id);
+      return (
+        UserInput == singleReservation.id ||
+        UserInput == singleReservation.firstName ||
+        UserInput == singleReservation.surname
+      );
+    });
+    this.setState({ reservations: myResult });
```
40thieves: This is great! Updating the search results on input is a good idea.

src/components/Results.js
```diff
@@ -12,39 +12,93 @@ const diffDays = (checkInDate, checkOutDate) => {
...
-      ))}
-    </table>
-  );
-};
+export default class Results extends Component {
+  constructor(props) {
+    console.log("receiving props");
+    console.log(props);
+    super(props);
+    this.state = { reservations: props.reservations };
```
40thieves: This code is correctly using state for the reservations, however I think state is actually unnecessary here. Remember in class how we talked about how we generally prefer to use props, until we **know** that the component needs to update this data when the interacts with it.

Here we are just using the reservations data directly - we aren't calling `this.setState`, so I think props is the better way of doing it.

src/containers/Bookings.js
40thieves: Now that we are actually using state to update the results, I think this is slightly incorrect. If you type something in the search input, it will filter the results correctly, but if you then click the search button, it will reset the results back to showing everything. I don't think that is the intended functionality.

However, I've realised that we're going to cover a better way of handling forms/input in the class today (called *refs*) that will hopefully make this easier to do.

PR #41, opened 2018-09-19
Cyf hotel react/elias

Eliasadam:
I have completed successfully home-work react leasson1. please review it. thanks in advance


src/components/Results.js
```diff
@@ -1,46 +1,73 @@
...
-
-  )
+        </thead>
+        <tbody>
+          {this.state.results.map(result => {
+
+            return (
+              <tr 
+                onClick={this.handleClick}
+                className={this.state.isRowToggled ? "onClick" : null}
```
40thieves: I think we discussed in class last week the solution to the problem where the toggle state is applied to all rows not just the one you click on. To recap: you need to extract each row as it's own component. Then each of these row components can hold it's own individual state.

src/containers/Bookings.js
```diff
@@ -2,28 +2,89 @@ import React, { Component } from 'react';
...
+    Counter:Results,
+    customerIdInput: "",
+    customerNameInput:"",
+   
+   
+  }
+
+  customerInputChange = (event) => {
+    this.setState({
+      customerIdInput: event.target.value
```
40thieves: We discussed in class that this uses the *controlled component* pattern, and how that's slightly unusual for this particular use. We looked at the *uncontrolled component* pattern too - where you use refs to capture the value of the input when clicking the button.

src/containers/Bookings.js
```diff
@@ -2,28 +2,89 @@ import React, { Component } from 'react';
...
+    this.setState({
+      results: results.filter(customer => customer.id === parseInt(customerIdInput, 10))
+    })
+  }
+  filteredCustomerName = (event) => {
+    const { customerNameInput, results } = this.state;
+    this.setState({
+      results: results.filter(customer => 
+          customer.firstName === customerNameInput || 
+          customer.surname === customerNameInput
```
40thieves: This is great! I really like the clear naming of variables.

src/containers/Bookings.js
```diff
@@ -2,28 +2,89 @@ import React, { Component } from 'react';
...
+  
+ 
 
+  filteredCustomerId = (event) => {
+    const { customerIdInput, results } = this.state;
+    // const customerIdInput = this.state.customerIdInput;
+    // const results = this.state.results;
+
+    this.setState({
+      results: results.filter(customer => customer.id === parseInt(customerIdInput, 10))
```
40thieves: Good usage of `parseInt` 👍 And adding the second argument because that is always necessary in JS.

PR #42, opened 2018-09-23
Week2 2 react

daleel1:


src/components/Search.js
```diff
@@ -1,22 +1,44 @@
...
+export default class Search extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: ""
+    };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
```
40thieves: Remember in class how we talked about using this syntax:

```js
handleChange = (event) => {
  // ...
}
```

If you use that, then you don't need to add the line in the constructor with `.bind(this)` in it.

src/components/Search.js
```diff
@@ -1,22 +1,44 @@
...
+    this.setState({ value: event.target.value });
```
40thieves: Because you're listening for updates on the `<input>` and updating the state, this pattern that you're using is called a "controlled component". There's not really a problem with using this pattern, but it is slightly unusual for this particular case. It is often used if we need to transform or keep track of exactly what the user has typed.

React has an alternative pattern (called "uncontrolled component"), were we don't add a onChange handler, and then just get the input when we click the button.

I've just realised that we haven't looked a feature of React that will make this easier in class yet (called *refs*). Hopefully that will make more sense on how we can do that after the class.

src/containers/Bookings.js
```diff
@@ -1,19 +1,37 @@
...
+// import FakeBookings from '../data/fakeBookings.json';
+import allBookings from "../data/fakeBookings.json";
 export default class Bookings extends Component {
-  search = () => {
-    console.info('TO DO!');
+  updateBookingsToDisplay = (input) => {
+    this.setState({ bookingsToDisplay: allBookings.filter(x => x.id === parseInt(input, 10))})
   };
-  
+  updateBookingsToDisplay2 =(input) =>{
```
40thieves: To make it easier for people to read and understand what this function is doing, I would rename it to something like `updateBookingsToDisplayByName`, and the function above to something like `updateBookingsToDisplayById`.

reber12: thanks a lot @40thieves for reviewing my PR. Really we can not thank you enough for your efforts.

PR #45, opened 2019-04-08
Kadir-CYF-Hotel

laperla11:
first commit


nbogie:
Review in progress.

src/Button.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+const Button = (props) => {
```
nbogie: Good, Kadir.
Note that with arrow functions where you're simply returning a single containing JSX element, you can use implicit return and also omit the braces, which makes things a bit more readable.  E.g.: 

```
const MyComponent = () =>     
  <div>
    <Greeting />
    <Mentor />
  </div>
```

laperla11:
Thank you very much for your time and feedback Neill.
I impelented your feedback.

Ahmet-K:
Look very nice and tidy. It is fabulous.

src/Orders.js
```diff
@@ -0,0 +1,33 @@
...
+
+class Orders extends Component {
+    constructor(props) {
+        super(props);
+        this.state = { orders: 0 }
+    }
+
+    addOrder = () => {
+        const previousCount = this.state.orders;
+        this.setState({ orders: previousCount + 1 });
```
40thieves: This isn't quite right. It will work _most_ of the time but if you click very quickly on the button you may notice that it will occasionally miss incrementing the number for a click.

This is because React will sometimes _batch_ multiple calls to `this.setState()` (this means that it will combine them together). This means that `previousCount` here will be wrong. To get around this, you must use the callback version of `setState`:

```js
this.setState((previousState) => {
  // ...
})
```

Can you fill out the rest?

The rule of thumb about which version of `setState` to use is that if you need to calculate the new state based from the previous state, then you need to use the callback version. If you don't need that then the simple object version is fine.

src/SearchResults.js
```diff
@@ -1,43 +1,63 @@
...
+                return { [result.id]: false };
+            })
+        }
+    }
+
+    handleClick = (event) => {
+        const clickedRow = event.target.parentElement;
+        const clickedRowIndex = clickedRow.rowIndex;
+        const previousState = this.state.isRowSelected[clickedRowIndex - 1][clickedRowIndex];
+        clickedRow.className = previousState ? '' : 'selected';
```
40thieves: This is something that you have to be careful about in React: you're directly manipulating the DOM. This is potentially a problem because now React doesn't know that your change is deliberate and will wipe away your change next time your render. Your manual DOM change is in the "real DOM", but not in React's virtual DOM and so it removes your manual change.

src/Orders.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    removeOrder = () => {
```
40thieves: Forgot to say: really nice job adding your own functionality here 👍

src/TableRow.js
```diff
@@ -0,0 +1,36 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            isRowSelected: false
+        }
+    }
+
+    handleClick = (event) => {
+        const clickedRow = event.target.parentElement;
+        const previousState = this.state.isRowSelected; clickedRow.className = previousState ? '' : 'selected';
```
40thieves: Just realised I posted a comment on some outdated code. Please ignore that comment.

This is something that you have to be careful about in React: you're directly manipulating the DOM. This is potentially a problem because now React doesn't know that your change is deliberate and will wipe away your change next time your render. Your manual DOM change is in the "real DOM", but not in React's virtual DOM and so it removes your manual change.

You've got 99% of the way there by splitting out a `TableRow` component. If you instead use the `selected` state in your `render()` method to add/remove this class then you'll be good! Well done figuring this problem out - it's intentionally quite tricky!

laperla11: Thanks for your time and feedback Ali, I will implement your feedback.

src/Clock.js
```diff
@@ -1,17 +1,20 @@
...
   componentDidMount() {
     // There is a bug in this component, where this.timer is not cleared if the
     // component is unmounted. It will continue to try to render if even the
     // component is no longer in the DOM.
-    this.timer = setTimeout(this.tick, 1000);
+    this.timer = setInterval(this.tick, 1000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
```
40thieves: Nice job!

src/Bookings.js
```diff
@@ -1,21 +1,117 @@
...
+  }
+
+  displayAllBookings = () => {
+    this.setState({ searchVal: " " });
+  };
+
   search = searchVal => {
-    console.info("TO DO!", searchVal);
+    this.setState({ searchVal: searchVal });
+    this.setState({
```
40thieves: You can combine these two `setState` calls together:

```
this.setState({
  searchVal: searchVal,
  searchResults: ...
})

File: src/Bookings.js [hunk 8]
```diff
@@ -1,21 +1,117 @@
...
+          <div className="container">
+            <Search
+              search={this.search}
+              displayAllBookings={this.displayAllBookings}
+            />
+            <SearchResults
+              results={
+                this.state.searchVal
+                  ? this.state.searchResults
+                  : this.state.fakeBookings
```
40thieves: If you changed a few things slightly it would make your component a bit simpler.

When we fetch the data, we store the full list of bookings. Then here we do the filtering based on `this.state.searchVal` (instead of in the `search` method). That means that if we change `this.state.searchVal` the bookings will be re-filtered.

So you `displayAllBookings` method just has to set `searchVal` to be an empty string and it will re-filter to show all results.

src/Bookings.js
```diff
@@ -1,21 +1,117 @@
...
+        // console.log(data);
+        this.setState({
+          isLoading: false,
+          fakeBookings: data
+        });
+      })
+      .catch(err => {
+        this.setState({
+          isLoading: false,
+          error: err
```
40thieves: Excellent work!

src/Bookings.js
```diff
@@ -1,21 +1,117 @@
...
+    this.setState({
+      searchResults: this.state.fakeBookings.filter(booking => {
+        return booking.firstName === searchVal || booking.surname === searchVal;
+      })
+    });
+  };
+
+  addBooking = newBooking => {
+    this.setState({
+      fakeBookings: this.state.fakeBookings.concat(newBooking)
```
40thieves: Really nice :)

src/TableHeader.js
```diff
@@ -0,0 +1,23 @@
...
+export default class TableHeader extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isColumnSorted: false
+    };
+  }
+
+  handleClick = e => {
+    this.props.sortColumn(this.props.columnKey, this.state.isColumnSorted);
```
40thieves: This is a really neat implementation! Well done :)

src/SearchResults.js
```diff
@@ -0,0 +1,73 @@
...
+        });
+      }
+    } else {
+      if (isColumnSorted) {
+        this.setState({
+          results: this.props.results.sort((a, b) => b[sortBy] - a[sortBy])
+        });
+      } else {
+        this.setState({
+          results: this.props.results.sort((a, b) => a[sortBy] - b[sortBy])
```
40thieves: Great job with the stretch goals!

laperla11:
Thanks for your kind words and the feedback Ali. I implemented your corrections.

PR #46, opened 2019-04-08
Ivina - React - Week 1

ivinapontes:


Gbaja:
Hi @ivinapontes. Well done on completing the exercise  🎉👍🏾. Couldn't really find any fault but some tips below:

- You used functional component throughout the app except for `App.js` , it doesn't seem to be doing much except for rendering jsx. Its good practise to use functional components whenever you can.

- Also a useful tip for working with objects in javascript. You can look into [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring). For example in your `Footer.js`, you can get the footer array from the argument directly as shown below:

```javascript
const Footer = ({ footerArray }) => {
  return (
    <div className="footer">
      <ul>
        {footerArray.map((footerItems, index) => (
          <li key={index}>{footerItems}</li>
        ))}
      </ul>
    </div>
  );
};
```

- Also, dont forget to take out comments.

ivinapontes:
Hey @Gbaja , thank you for your feedback. I appreciate it!
I'll fix my code following the suggestions you gave me! :smile_cat: :tada:

src/Highlight.js
```diff
@@ -0,0 +1,46 @@
...
+        highlighted: !previousState.highlighted
+      };
+    });
+  }
+
+  render() {
+    return (
+      <tr
+        className={this.state.highlighted ? "highlighted" : " "}
+        onClick={() => this.changeColor()}
```
nbogie: can be simplified to `onClick={this.changeColor}`

src/Highlight.js
```diff
@@ -0,0 +1,46 @@
...
+      >
+        <td>{this.props.customer.id}</td>
```
nbogie: There are some advantages (as you go forward) to creating these `td`s by mapping through a list of the keys.  However, your approach is the currently very readable.  

You could also temporarily assign `this.props.customer` to another variable, to reduce the amount of `noise`.
`const cust = this.props.customer` at the start of `render()`, before the `return` statement begins.  That would allow you to then write:
```
<td>{cust.title}</td>
<td>{cust.firstName}</td>
```
 etc

PR #47, opened 2019-04-09
Mohammad - hotel

markazie:


src/components/Button.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
```
40thieves: Great job organising this into a `components` directory!

src/App.js
```diff
@@ -1,16 +1,30 @@
...
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">CYF Hotel</header>
+        <Heading />
         <Bookings />
         <Restaurant />
+        <div className="card">
+          <TouristInfoCards src="1" alt="London" city="London" />
```
40thieves: This isn't really wrong, but the name `TouristInfoCards` implies a *plural* (because of the "s" on the end) but only 1 card is being shown.

I would rename your existing component to be called `TouristInfoCard` (notice there's no "s") and then create a new component `TouristInfoCards` which has the `<div className="card">` and all 3 cards inside it.

src/components/Button.js
```diff
@@ -0,0 +1,7 @@
...
+
+export default function Button() {
+  return (
+          <button className="btn btn-primary">Search IDs</button>
```
40thieves: This component is not very reusable right now - if we tried to use it elsewhere in our app then it would always say "Search IDs". How could we change this component to be more reusable?

src/components/Footer.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+
+export default function Footer(props) {
+    const li = props.data.map((data, index) => <li key={index}>{data}</li>)
```
40thieves: I actually quite like following the pattern you've got here, but most people who write React would normally *inline* the `li` variable within the curly braces.

I think sometimes it's useful to separate them out like you have though.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export default function TouristInfoCards(props) {
+  return (
+    <div class="">
```
40thieves: Remember that we have to use `className` when writing React

src/App.js
```diff
@@ -1,16 +1,30 @@
 import React, { Component } from "react";
-
+import Heading from './components/Heading'
 import Bookings from "./Bookings";
```
nbogie: Why are some components in `src/components/` and others in `src/`?  Your reader (your coworkers, or future you) will waste time thinking about this, and will waste time trying to remember which components are where.
Be consistent.

src/Restaurant.js
```diff
@@ -1,17 +1,16 @@
...
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
-      </ul>
-    </div>
-  );
-};
-
-export default Restaurant;
+  render() {
```
nbogie: You can convert this back into a functional component now.  It has no state.

src/components/Orders.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  addOrder = () =>
+    this.setState(preState => {
+      return { orders: preState.orders + 1 };
+    });
+
+  render() {
+    return (
+      <li>
```
nbogie: good job

src/components/SearchResults.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from "react";
+import moment from "moment";
+
+export default class SearchResults extends Component {
+  constructor() {
+    super();
+    this.state = {};
```
nbogie: This initial state should have a `selected` property, if you're going to say the following later:
```return { selected: !pre.selected };```

However, I'm not sure that having all of SearchResults have a selected flag makes any sense - we want to track selection of specific rows.

File: src/components/SearchResults.js [hunk 10]
```diff
@@ -0,0 +1,41 @@
...
+      : (event.target.parentNode.className = "");
+  };
+
+  render() {
+    const result = this.props.results.map(r => (
+      <tr key={r.id} data-id={r.id} onClick={this.rowSelected}>
+        <th scope="row">{r.id}</th>
+        <td>{`${r.title} ${r.firstName} ${r.surname}`}</td>
+        <td>{r.email}</td>
+        <td>{`room: ${r.roomId}`}</td>
```
nbogie: which is simpler?
```<td>{`room: ${r.roomId}`}</td>```
```<td>room: {r.roomId}</td>```

File: src/components/SearchResults.js [hunk 11]
```diff
@@ -0,0 +1,41 @@
...
+    this.state.selected
+      ? (event.target.parentNode.className = "bg-warning")
+      : (event.target.parentNode.className = "");
+  };
+
+  render() {
+    const result = this.props.results.map(r => (
+      <tr key={r.id} data-id={r.id} onClick={this.rowSelected}>
+        <th scope="row">{r.id}</th>
+        <td>{`${r.title} ${r.firstName} ${r.surname}`}</td>
```
nbogie: which is simpler?
```<td>{`${r.title} ${r.firstName} ${r.surname}`}</td>```
```<td>{r.title} {r.firstName} {r.surname}</td>```

File: src/components/SearchResults.js [hunk 12]
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from "react";
+import moment from "moment";
+
+export default class SearchResults extends Component {
+  constructor() {
+    super();
+    this.state = {};
+  }
+
+  rowSelected = event => {
```
nbogie: I think this will simply select and deselect ALL rows any time ANY row is clicked.  Not what was specified in the HW task.

src/components/SearchResults.js
```diff
@@ -0,0 +1,41 @@
...
+    this.setState(pre => {
+      return { selected: !pre.selected };
+    });
+
+    this.state.selected
```
nbogie: You don't want to directly modify the DOM in this way. Ali will explain why tomorrow in class.

Instead, modify your component's state, and in the render method, build the DOM according to the state.

PR #49, opened 2019-04-09
Elamin CYF Hotel

aldeez988:


clems27:
Good job, nice use of props, well-done @aldeez988

src/TouristInfoCards.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Cards from './Cards';
+import faker from 'faker';
+const TouristInfoCards = () => {
+    return (
+        <div className="card-deck">
+            <Cards website="https://peoplemakeglasgow.com/" city="Glasgow" image={faker.image.city()} />
```
40thieves: Huh, that's cool - I didn't realise Faker could provide images too

src/SearchButton.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+
+const SearchButton = () => {
+    return (
+
+        <button className="btn btn-primary">Search IDs</button>
```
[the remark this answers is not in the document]
aldeez988: We can change the button label and send it as ac props from the parent component also I need to change the component name from SearchButton to Button

src/TouristInfoCards.js
aldeez988: Thanks @40thieves  for you valuable feedback I have update my PR +1

src/Orders.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import RestaurantButton from './RestaurantButton';
+
+class Orders extends React.Component {
```
nbogie: good

src/Restaurant.js
```diff
@@ -1,17 +1,20 @@
...
-      <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
-      </ul>
-    </div>
-  );
+  render() {
+    const pizzas = 0;
```
nbogie: remove unused code.

src/SearchResults.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import moment from 'moment'
+
+class SearchResults extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = { isRowSelected: false, selectedRow: null };
+    }
+
```
nbogie: if you read HW task 14 carefully, you'll see that many rows can be selected at the same time.  So you need a data structure(s) that can track that.  A single scalar variable (e.g. a single number) isn't going to be enough to model the case that the rows 1, 2, 5, 17, and 19 are selected!

Once you decide on the correct state to use, it will be easier to write everything else.

src/SearchResults.js
```diff
@@ -0,0 +1,80 @@
...
+            event.target.parentNode.classList.add("table-danger");
+
+        } else if (this.state.selectedRow === event.target.parentNode) {//here we check if the clicked row is previously selected  if it is it will be unhighlighted
+
+            this.setState({
+                isRowSelected: false,
+                selectedRow: null
+            })
+            // removing the added class in case we clicked a previously highlighted row 
+            event.target.parentNode.classList.remove("table-danger");
```
nbogie: Don't do direct manipulation of the DOM outwith the render method (where you BUILD the DOM). Instead, change the STATE (or props of a child) and let the render() method(s) build the DOM for you.

aldeez988: Thank you @nbogie for your feedback, I have updated my code

src/Components/SearchResults.js
```diff
@@ -0,0 +1,142 @@
...
+        this.setState({ results: nextProps.results });
+    }
+
+    sortAndUpdate = (nameOFCoulumn, clicked) => {
+        console.log(nameOFCoulumn)
+        const ascendingSort = (a, b) => {
+            console.log(a.firstName)
+            if (nameOFCoulumn === "First Name") {
+                if (a.firstName < b.firstName) { return -1; }
+                if (a.firstName > b.firstName) { return 1; }
```
40thieves: This method repeats itself a bit. How could we clean it up so that it is more DRY?

PR #50, opened 2019-04-10
Clement CYF-HOTEL-REACT

clems27:


aldeez988:
Well done nice work @clems27

src/App.js
```diff
@@ -2,18 +2,30 @@ import React, { Component } from "react";
...
+import Footer from "./Footer"
 import "./App.css";
+import "./grid.css";
 
 class App extends Component {
   render() {
-    return (
-      <div className="App">
-        <header className="App-header">CYF Hotel</header>
+    const contactInfo = [
```
40thieves: Nice job with naming 👏

src/Button.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const Button = (props) => {
+  return <button className="btn btn-primary">{props.text}</button>;
```
40thieves: Great job making the `Button` component reusable!

src/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <footer className="footer">
+      <ul>
+        {props.contactInfo.map((info, index) => {
+          return <li key={index}>{info}</li>;
```
40thieves: Great usage of the `key`!

src/Restaurant.js
```diff
@@ -1,13 +1,13 @@
...
-const Restaurant = () => {
+const Restaurant = (props) => {
   const pizzas = 0;
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
         <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
+          Pizzas: {pizzas} <button className="btn btn-primary">{props.text}</button>
```
40thieves: Could we reuse the `<Button>` component here?

src/App.js
```diff
@@ -2,18 +2,30 @@ import React, { Component } from "react";
...
+      "0123 456789"
+    ];
+   return (
+    <div className="App">
+      <div className="container">
+        <Heading />
+        <TouristInfoCards />
         <Bookings />
-        <Restaurant />
+        <Restaurant text={"add"}/>
```
40thieves: This isn't necessarily incorrect, but it is a little unusual. We can't have to pass down props for every bit of data in our application - our `App` component would be very complex then!

In this case, I would say that the `App` component "doesn't care" about what text the button has inside `Restaurant`. That's the job of the `Restaurant` component. This is sort of what I meant when I was talking about how components should have 1 *responsibility*

src/Search.js
```diff
@@ -17,7 +18,7 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer Id"
               />
-              <button className="btn btn-primary">Search IDs</button>
+              <Button text={"Search IDs"}/>
```
40thieves: This isn't incorrect, but if you're just passing a string as a prop, then the curly braces are unnecessary: `<Button text="Search IDs: />` works fine too. And it's 2 less characters to type :)

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+              <tr>
+                <th scope="row">{result.id}</th>
+                <td>{result.title}</td>
+                <td>{result.firstName}</td>
+                <td>{result.surname}</td>
+                <td>{result.email}</td>
+                <td>{result.roomId}</td>
+                <td>{result.checkInDate}</td>
+                <td>{result.checkOutDate}</td>
+                <td>{moment(result.checkOutDate).diff(result.checkInDate, 'days')}</td>
```
40thieves: Great job!

src/TouristInfoCards.js
```diff
@@ -0,0 +1,82 @@
...
+                South Bank cultural complex, and the entire city.
+                </p>
+                <a href="https://www.visitlondon.com/" class="btn btn-primary">
+                Chekout Upcoming events
+                </a>
+            </div>
+        </div>
+
+        <div class="card">
+            <img
```
40thieves: There's some repeated code in this component. How could we improve this component so that we're not repeating ourselves so much?

src/App.js
clems27: Thanks

src/Restaurant.js
clems27: I have updated the code to use the button component. thanks for pointing that out

src/App.js
clems27: Looking back at my code, I realise there is a better way to do this, will be updating this line shortly

src/TouristInfoCards.js
clems27: I figured this will be the best approach considering the way I want to render the page in general

src/SearchResults.js
```diff
@@ -0,0 +1,67 @@
...
+
+    };
+  }
+  selection = (props) => {
+    this.setState(previousState => {
+      return {selected: !previousState.selected}
+    });
+    this.state.selected
+    ? (props.target.parentNode.className="highlight")
+        : (props.target.parentNode.className="" );
```
40thieves: This is something that you have to be careful about in React: you're directly manipulating the DOM. This is potentially a problem because now React doesn't know that your change is deliberate and will wipe away your change next time your render. Your manual DOM change is in the "real DOM", but not in React's virtual DOM and so it removes your manual change.

src/Clock.js
```diff
@@ -8,19 +8,20 @@ class Clock extends Component {
...
 
   componentDidMount() {
-    // There is a bug in this component, where this.timer is not cleared if the
-    // component is unmounted. It will continue to try to render if even the
-    // component is no longer in the DOM.
-    this.timer = setTimeout(this.tick, 1000);
+      this.timer = setInterval(this.tick, 1000);
+    }
+    componentWillUnmount() {
+      clearInterval(this.timer);
```
40thieves: Well done!

src/Search.js
```diff
@@ -1,29 +1,45 @@
...
-              />
-              <button className="btn btn-primary">Search IDs</button>
-            </div>
-          </form>
+class Search extends Component {
+  setInput = input => {
+    this.input = input;
+  };
+  handleSubmit = event => {
+    this.props.search( this.input.value)
```
40thieves: 👍

src/Bookings.js
```diff
@@ -1,21 +1,95 @@
...
+        this.setState({
+          isLoading: false,
+          error: error
+        });
+      });
+  }
+addBooking =(newBooking)=>{
+ // console.log(newBooking)
+this.setState({
+  searchResult:[...this.state.searchResult, newBooking]
```
40thieves: Nice use of the spread (`...`) operator!

src/Bookings.js
```diff
@@ -1,21 +1,95 @@
...
+          return res;
+        } else {
+          throw new Error("HTTP error");
+        }
+      })
+      .then(res => res.json())
+      .then(data => {
+        this.setState({
+          Bookings: data,
+          searchResult: data,
```
40thieves: We generally say in React that it is better to calculate things when we need them rather than try to store everything in state. Here we only need to store the current value and the full list of results to be able to calculate the filtered list of results.

If we added a button to show all results again (after filtering), how would you implement that?

Hint: it might be easier to store just the current `searchVal` in state, then calculate the filtered results in the render method. Then if you want to show all the results again, you can just set the `searchVal` to be empty and then the filter will return all the results.

PR #51, opened 2019-04-10
Matthew React Week Three

mtbdata711:


src/CustomerDataTable.js
```diff
@@ -0,0 +1,36 @@
...
+            </th>
+            <td key={res.title}>{res.title}</td>
+            <td key={res.firstName}>{res.firstName}</td>
+            <td key={res.surname}>{res.surname}</td>
+            <td key={res.email}>{res.email}</td>
+            <td key={"roomId"}>{`00${index + 1}`}</td>
+            <td key={res.checkInDate}>{res.checkInDate}</td>
+            <td key={res.checkOutDate}>{res.checkOutDate}</td>
+            <td key={"nights"}>
+              {calculateStayDuration(res.checkInDate, res.checkOutDate)}
```
clems27: Well done Matthew, I really like the way you use your table.

src/PrimaryButton.js
```diff
@@ -0,0 +1,15 @@
...
+const PrimaryButton = (props) => {
+
+    return (
+        <div className="primary-button">
+            <button className="btn btn-primary">
+                {props.buttonText}
+            </button>
+        </div>
+    )
+}
```
clems27: Good job Matthew, good to see you use props here, welldone

src/CustomerDataTable.js
```diff
@@ -0,0 +1,36 @@
...
+function calculateStayDuration(checkIn, checkOut) {
+  return moment(checkOut).diff(moment(checkIn), "days");
+}
+
+const CustomerDataTable = customerObj => {
+  return (
+    <tbody>
+      {customerObj.results.map((res, index) => {
+        return (
+          <tr key={`tr${index}`}>
```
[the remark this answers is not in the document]
nbogie: You need only put a key on the top containing node.  So in this case: 
```<tr key={res.id}>```.  None of the other child nodes need a key.

File: src/CustomerDataTable.js [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+            <th scope="row" key={res.id}>
+              {res.id}
+            </th>
+            <td key={res.title}>{res.title}</td>
+            <td key={res.firstName}>{res.firstName}</td>
+            <td key={res.surname}>{res.surname}</td>
+            <td key={res.email}>{res.email}</td>
+            <td key={"roomId"}>{`00${index + 1}`}</td>
```
nbogie: Just read ```res.roomId``` from the result?

src/PrimaryButton.js
nbogie: Good stuff!

src/TableHeadings.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+
+
+const TableHeadings = (props) => {
+    return (
```
nbogie: These arrow functions can be written more simply with implicit return and no braces delimiting the body (because you just have one statement (one parent JSX node creation)): 
```const TableHeadings = props => <thead>
<tr>...
</thead>
```

There's nothing wrong with the way you've done it, however.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div className="tourist-information">
+      <div className="container">
+        <div className="row">
+          {props.data.map((singleCard, index) => {
```
nbogie: For me this is crying out for a `TouristInfoCard` component.  You could define it within this same file.

nbogie: Good work driving this by looping over data.

src/DisplayBookings.js
```diff
@@ -0,0 +1,29 @@
...
+            "First Name",
+            "Surname",
+            "Email",
+            "Room ID",
+            "Check In Date",
+            "Check Out Date",
+            "Nights"
+          ]}
+        />
+        {<CustomerDataTable results={FakeBookings} />}
```
40thieves: I think these curly braces are unnecessary here :)

src/CustomerDataTable.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+
+toggleHighlightedRow = (e) => {
+  this.setState(prevState => ({
+    selected: !prevState.selected
+  })
+)
+this.state.selected ? (e.target.parentNode.className = 'bg-info') : (e.target.parentNode.className = '');
```
40thieves: This is something that you have to be careful about in React: you're directly manipulating the DOM. This is potentially a problem because now React doesn't know that your change is deliberate and will wipe away your change next time your render (anywhere in the app). Your manual DOM change is in the "real DOM", but not in React's virtual DOM and so it removes your manual change.

src/Orders.js
```diff
@@ -0,0 +1,36 @@
...
+  super(props);
+  this.state = {
+    orders: 0
+  };
+}
+
+addOrder = () => {
+  this.setState(previousState => {
+    return {
+      orders: previousState.orders + 1
```
40thieves: Great job!

src/Restaurant.js
```diff
@@ -1,17 +1,23 @@
...
+
+
+  render() {
+        return (
+      <div>
+        <h3>Restaurant Orders</h3>
+        <ul>
+          <Orders orderType="Salads" />
+          <br/>
+          <Orders orderType="Pizzas" />
```
40thieves: Can you see how making our components reusable (but configurable via props) is useful? :)

src/CustomerDataTable.js
mtbdata711: Hey Ali! Thanks for this. Your commitments got me really keen to re-visit this and I've now added a revision with a new component. Let me know if this was along the right track.

src/Clock.js
```diff
@@ -14,6 +14,10 @@ class Clock extends Component {
     this.timer = setTimeout(this.tick, 1000);
   }
 
+  componentWillUnmount(){
+    clearTimeout(this.timer);
```
40thieves: Great! :)

src/Bookings.js
```diff
@@ -1,19 +1,77 @@
...
 import Search from "./Search.js";
-// import SearchResults from './SearchResults.js';
-// import FakeBookings from '../data/fakeBookings.json';
+import SearchResults from "./SearchResults.js";
 
 export default class Bookings extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      FakeBookings: null,
```
40thieves: This is a tiny nitpick, but generally we prefer camelCase variable naming for state variables.

src/Bookings.js
```diff
@@ -1,19 +1,77 @@
...
 
   render() {
     return (
       <div className="App-content">
         <div className="container">
+          {this.state.err && this.errorMessage()}
           <Search search={this.search} />
-          {/* <SearchResults results={FakeBookings} /> */}
+
+          {this.state.isLoading ? (
```
40thieves: I believe if there was an error, it would continue to show a loading state?

src/Bookings.js
```diff
@@ -1,19 +1,77 @@
...
+    } else {
+      this.setState({ SearchResults: filteredSearch });
+    }
+  };
+
+  loadingMessage = () => {
+    return <span>Loading...</span>;
+  };
+
+  errorMessage = () => {
```
40thieves: We haven't talked about it much it class but for this case I would prefer to put this in it's own little component. It shows up nicer in then dev tools then.

Usually if I'm making something like this where the component will only be shown in this one file, that's where I break the rule about how every component should be in it's own file.

src/Bookings.js
```diff
@@ -1,19 +1,77 @@
...
+        } else {
+          throw new Error("HTTP error");
+        }
+      })
+      .then(res => res.json())
+      .then(data =>
+        this.setState({
+          FakeBookings: data,
+          isLoading: false,
+          SearchResults: data
```
40thieves: We generally say in React that it is better to calculate things when we need them rather than try to store everything in state. Here we only need to store the current value and the full list of results to be able to calculate the filtered list of results.

If we added a button to show all results again (after filtering), how would you implement that?

Hint: it might be easier to store just the current `searchVal` in state, then calculate the filtered results in the render method. Then if you want to show all the results again, you can just set the `searchVal` to be empty and then the filter will return all the results.

src/Search.js
```diff
@@ -1,29 +1,47 @@
...
+
+class Search extends Component {
+
+input = (inputEl) => {
+    this.inputRef = inputEl
+}
+
+  handleSubmit = event => {
+    event.preventDefault()
+    return this.props.search(this.inputRef.value);
```
40thieves: Well done :)

PR #52, opened 2019-04-11
Madiha React Week 1 

diya2214:


src/App.js
```diff
@@ -1,19 +1,47 @@
...
 class App extends Component {
-  render() {
+  render () {
     return (
-      <div className="App">
-        <header className="App-header">CYF Hotel</header>
+      <div className=' App'>
+        <Heading />
+        <div className='Card-row'>
+          <TouristInfoCards
```
40thieves: This component name implies that it renders multiple cards but it actually just renders one. Which isn't really a problem but we can make it more readable.

I would rename your existing component to `TouristInfoCard` (no "s" at the end) and then create a new component named `TouristInfoCards` which then renders the 3 cards that you have here.

src/Bookings.js
```diff
@@ -1,21 +1,21 @@
...
-  render() {
+  render () {
     return (
-      <div className="App-content">
-        <div className="container">
+      <div className='App-content'>
+        <div className='container'>
           <Search search={this.search} />
-          {/* <SearchResults results={FakeBookings} /> */}
+          {<SearchResults results={FakeBookings} />}
```
40thieves: You don't actually need the curly braces here. The reason that they were here before is because we needed to switch to "normal JS" to put the `<SearchResults>` component in a comment. Now that we're not commenting it out the curly braces are unnecessary

src/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+const Footer = prop => {
+  return (
+    <ul className='footer'>
+      {prop.value.map((addressLine, index) => {
+        return <li key={index}>{addressLine}</li>
```
40thieves: Nice naming :)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+const TouristInfoCards = props => {
+  return (
+    <div class='card' style={{ width: '18rem', right: '4' }}>
```
40thieves: Remember that when you're writing a React component you have to convert `class` to `className`

src/Bookings.js
diya2214: noted . thanks

src/App.js
```diff
@@ -1,19 +1,47 @@
...
...CSXfv8AyaCxy/qxVfJQJ++PvVf8QXES33GbVI3PSuL/ABRgd4mZMFmIGSNmnGKpuPJbp7dakst9iR6dR5ZB/eV394n+vKprPMiTDkkeYB9juKDFk+FSJYqibLOEPQ0nCc0X5lkkBVkiDgdNO3rUHM+csyxiDic/yqv4bh8bVK3DyoXwJPvH/iKlyfBdaxRTGaQTyq2Xl39YqReA8v1q6xSYt5UiusIcZNEG0aPtcNG4qfsRVv6dsCzpIztq6O2Pd/e6+XlXFm7+zfA/B0J6+dKlSDQPeuNptR/FtjqPCjOEBPEydmxM+IUGlSqTXhYYPxr4ms4lS4Kj5px19In6e9GcFpQmY1qAHadwBD5PyJMk9TSpVzI8x1OnPiVmd5xwhtXTGAe+mIwdxHT0PQiqu9bKstwZGGHU4Ox8SCI+lKlXRwTcscWznZopTaRsLDzyoFp739qLRB+Zpx0rnkdqy+pLqF7ZuOSJK94aSMrmKVKkwXL/APpjpeXwX0L+1wnAaSo4RJB6FiQfEEmq1uStqi1blcwSUGDkgyR1HhSpV0JdPD2W3mc5dTP+ocPInPJLyBrjKgVRPzS0AbKApB96sf7MOzMvlEtBSuJd3vDBMEEMRPoaVKuL1a1dI7fSSbVlJxvLWS+e9qtroYd6TqCEGfDJ+1VFzhnJ1IT8+pvZhjzzT0qPTSbxq/cLzqpsL4K0NLrvgBp8J++9WHJblp7pDns4QLKR3mlAusNjI3iM0qVbcP8AcZMvYLfiLtm6zI2qYmQQTAOZEwQOs+sjFVXMeCFw61QW7u5A7ttzMggzFt/+0+VNSrsS6eEkm1ycNdXkhPVdv9lXa4jQyJcSAjsWBBDSQu89e6PvVdzm2XSypMzdM9NwBv70qVcqUUnwdiMm1yTfBRYXj3iBouR5N3RJHoTV/wDD9rvLdYgr+1kHoNKhSD/mDj3pUqw55yTde5fU3dNCMoq/K39DPc7v6uzVY7qaT0yXZs+ean1dnwdpDEXGvOM+BVf0+9KlWtxXhXv+zZlUv7n7vukS8n0ixf0H52tLB8EQAmD1LFp9KqOE4bMUqVGMa2f52Kzd6r87hvObSpeFsR3EtoY/eOtifefcVAeFpUqb0yuCbF9U6m6H/sgrteHpUq2qCMLmya3bFd9kKelQSRNm0LT41KGBH+9PSpl0CrO1tg/0KY2fX7UqVSwUf//Z'
```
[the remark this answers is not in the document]
diya2214: got it

src/Orders.js
```diff
@@ -0,0 +1,27 @@
...
+class Orders extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      orders: 0
+    }
+  }
+  addOrder = () => {
+    this.setState(previousState => {
+      return { orders: previousState.orders + 1 }
```
40thieves: 👍

src/Restaurant.js
```diff
@@ -1,17 +1,14 @@
...
-export default Restaurant;
+import React, { Component } from 'react'
+import Orders from './Orders'
+class Restaurant extends Component {
+  render () {
+    return (
+      <div>
+        <h3>Restaurant Orders</h3>
+        <Orders orderType='Pizza' />
+        <Orders orderType='Salad' />
```
40thieves: Can you see how making our components reusable (but configurable via props) is helpful?

src/SearchResults.js
```diff
@@ -0,0 +1,79 @@
...
+        </thead>
+        <tbody>
+          {this.props.results.map(row => {
+            return (
+              <tr
+                id={row.id}
+                key={row.id}
+                onClick={this.toggle}
+                className={
+                  this.state.selectedRowId === row.id ? 'bg-info' : 'bg-light'
```
40thieves: Really great job! Well done figuring this out - it is quite a tricky problem :)

However there is actually a much simpler approach to this problem. Do you remember in class we had the example with multiple Counter components? Each counter remembered their own count.

We can do a similar thing here - if we extracted a table row component (one for each result), then they could each remember their own state. In fact they would only need to remember if they were selected or not (which sounds quite like a boolean to me 😉). How would we go about implementing that?

src/Orders.js
```diff
@@ -0,0 +1,27 @@
...
+    })
+  }
+  render () {
+    return (
+      <ul>
+        <li>
+          {this.props.orderType}: {this.state.orders}{' '}
+          <Button caption='Add' addOrder={this.addOrder} />
```
40thieves: Great job recognising that we had already implemented a `Restaurant` button component. Do you see how we're passing the `this.addOrder` function as a prop, which is then called by `Button` when it is clicked?

src/Clock.js
```diff
@@ -1,26 +1,24 @@
...
-  componentDidMount() {
-    // There is a bug in this component, where this.timer is not cleared if the
-    // component is unmounted. It will continue to try to render if even the
-    // component is no longer in the DOM.
-    this.timer = setTimeout(this.tick, 1000);
+  componentDidMount () {
+    this.timer = setInterval(this.tick, 1000)
+  }
+  componentWillUnmount () {
+    clearInterval(this.timer)
```
40thieves: Well done :)

src/Search.js
```diff
@@ -1,29 +1,44 @@
...
-              />
-              <button className="btn btn-primary">Search IDs</button>
-            </div>
-          </form>
+class Search extends Component {
+  setInputRef = input => {
+    this.input = input
+  }
+  submitHandler = e => {
+    this.props.search(this.input.value)
```
40thieves: 👍

src/Bookings.js
```diff
@@ -1,21 +1,83 @@
...
+import React, { Component } from 'react'
+import Search from './Search.js'
+import SearchResults from './SearchResults.js'
+import NewBooking from './NewBooking.js'
 
 export default class Bookings extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      FakeBookings: [],
```
40thieves: Little nitpick: generally we prefer camelCase naming for our state variables.

src/Bookings.js
```diff
@@ -1,21 +1,83 @@
...
-    console.info("TO DO!", searchVal);
-  };
-
-  render() {
-    return (
-      <div className="App-content">
-        <div className="container">
-          <Search search={this.search} />
-          {/* <SearchResults results={FakeBookings} /> */}
+    if (searchVal !== '' || searchVal !== null) {
```
40thieves: Well done spotting the problem here. There may be a slightly improved approach. We generally say in React that it is better to calculate things when we need them rather than try to store everything in state. Here we only need to store the current value and the full list of results to be able to calculate the filtered list of results.

It might be easier to store just the current `searchVal` in state, then calculate the filtered results in the render method. Then if you want to show all the results again, you can just set the `searchVal` to be empty and then the filter will return all the results.

src/Bookings.js
```diff
@@ -1,21 +1,83 @@
...
+        } else {
+          throw new Error('HTTP error')
+        }
+      })
+      .then(res => res.json())
+      .then(data => {
+        console.log(data)
+        this.setState({
+          isLoading: false,
+          FakeBookings: data
```
40thieves: Great!

src/Bookings.js
```diff
@@ -1,21 +1,83 @@
...
+      })
+    } else {
+      this.setState({
+        FakeBookings: this.state.FakeBookings
+      })
+    }
+  }
+  addBooking = newBooking => {
+    this.setState({
+      FakeBookings: this.state.FakeBookings.concat(newBooking)
```
40thieves: Well done with the stretch goal!!

src/Bookings.js
diya2214: I tried this with nested filter and map statement but when searchVal is ' ' or null filter return empty array means nothing to map where I want filter to return all the results ..... Can you explain a bit more your approach please

40thieves: I was thinking something like this: https://gist.github.com/40thieves/8f42634bdb49b9be5a31768a32645cc3

So yes you are correct that you need to handle empty strings or `null`, but my point was more about what you are storing in state.

In my approach, I'm storing **all** the results and the search string. Then we compute the currently filtered results in the `render` method. This is generally a simpler approach because we don't have to worry about keeping the state correct at all times. So for example if we added another search field (e.g. dates) then we'd have to make sure that we updated the state correctly to filter by name and by date. In my approach you would only need to update the component store the date that the user had entered  and update the `filterResults` method to also check that value in the filter.

This is why we generally say if it is possible to **compute** a value at render time that is the better approach.

Does that make more sense?

PR #53, opened 2019-04-12
Tayo-hotel-homework

Atayo:


ahmedahashim:
good work , if you create a separate file for TouristInfoCards instead of putting in App.js file ,is it makes sense?

src/App.js
```diff
@@ -2,17 +2,44 @@ import React, { Component } from "react";
...
-      <div className="App">
-        <header className="App-header">CYF Hotel</header>
-        <Bookings />
-        <Restaurant />
-      </div>
-    );
+		<div className="App">
+			<Heading />
+			<div className="App-TouristInfoCards">
+				<TouristInfoCards
```

src/Footer.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const Footer = props => {
+	console.log(props);
+	return (
+		<ul>
+			{props.footerData.map(name => {
+				return <li key>{name}</li>;
```
40thieves: Well done!

src/SearchButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function SearchButton() {
+  return (
+    <div>
+      <button className="btn btn-primary">Search IDs</button>
```
40thieves: How could we make this component more reusable? Right now it will always show "Search IDs" as the button text. Which isn't very useful if we wanted to use it somewhere else in the app. How could we change this so that it can be reused in multiple different places?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+
+const TouristInfoCards = props => {
+    return (
+        <div class="card" style={{ width: "18rem" }}>
```
40thieves: Remember that when writing a React application you have to convert `class` to `className`

PR #55, opened 2019-04-13
Ahmed-react-hotel-Lesson 1

ahmedahashim:


src/App.js
```diff
@@ -1,19 +1,32 @@
 import React, { Component } from "react";
-
+import TouristInfoCards from "./TouristInfoCards";
 import Bookings from "./Bookings";
 import Restaurant from "./Restaurant";
 import "./App.css";
-
+import Heading from "./Heading";
+import ReactDOM from "react-dom";
```
40thieves: The `ReactDOM` import is not necessary. If you look inside `src/index.js` you'll see that we already have set up the `ReactDOM` import & call for you. You don't need to worry about it.

In my examples that we looked at in class it is needed because the entire application is in 1 file

src/App.js
```diff
@@ -1,19 +1,32 @@
...
+import Footer from "./Footer"
 class App extends Component {
   render() {
     return (
       <div className="App">
         <header className="App-header">CYF Hotel</header>
```
40thieves: When *extracting* a component you should replace the previous component with the new component

src/Cards.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+const Cards = (props) => {
+  return (
+    <div class="card" style={{ width: "18rem" }}>
```
40thieves: Remember that you need to convert `class` to `className` when using React

src/App.js
```diff
@@ -1,19 +1,32 @@
...
+        <Heading />
+        
+        <TouristInfoCards />
         <Bookings />
         <Restaurant />
+
+        <Footer
+          Details={[
```
40thieves: Normally we name props with `camelCase` (lowercase letter first)

src/SearchBtn.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+const SearchBtn = () => {
+    return (
+        <button className="btn btn-primary">Search IDs</button>
```
40thieves: How could we make this component more reusable? Right now if we tried to use this component elsewhere in the app it would still say "Search IDs". How could we change it so that it can be used in multiple places?

src/Cards.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+const Cards = (props) => {
```
40thieves: The naming here implies that the component shows multiple cards, but it's only showing 1. I would normally call this `Card` instead.

src/App.js
ahmedahashim: done

src/Cards.js
ahmedahashim: done

src/App.js
ahmedahashim: done

src/App.js
ahmedahashim: done

PR #57, opened 2019-04-13
Alisina-Hotel-Proj

Alisina92:
I have done it

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,22 +12,23 @@ class App extends Component {
     return (
       <div className="App">
         <Heading />
-
-        <TouristInfoCards
-          image="https://i.ytimg.com/vi/c8Sa0l98UB4/maxresdefault.jpg"
-          city="London"
-          link="http:/www.visitlondon.com"
-        />
-        <TouristInfoCards
-          image="https://www.visitbritain.org/sites/default/files/styles/large/public/vb-corporate/Images/basic-page/top-attractions-500x200_0.jpg?itok=eA8FDRfQ"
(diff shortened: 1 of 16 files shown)


src/App.js
```diff
@@ -3,14 +3,40 @@ import React, { Component } from "react";
...
 
+import Footer from "./Footer";
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">CYF Hotel</header>
+        <Heading />
+
+        <TouristInfoCards
```
40thieves: This component name implies that it shows multiple cards but it only shows one. I would probably rename your existing component to `TouristInfoCard` (no "s" at the end) and then create another component named `TouristInfoCards` which renders all 3 of the cards.

src/Button.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const Button = () => {
+  return <button className="btn btn-primary">Search IDs</button>;
```
40thieves: Could we improve this component a bit by making it reusable? Right now if we tried to show this component elsewhere in the app then it would still show "Search IDs". How could we change this component so that it can be used in other parts of our app?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div class="card" style={{ width: "18rem" }}>
```
40thieves: Remember that when writing React components we have to convert them to use `className` not `class`

PR #58, opened 2019-04-13
Mussie-CYF-Hotel-Lesson1

Mussie14:


NajlaSoltana:
I like your design and the pictures you choose, especially the logo you've created. you are a designer!!

Mussie14:
final

src/App.js
```diff
@@ -1,16 +1,45 @@
-import React, { Component } from "react";
+import React, { Component } from 'react';
+
+import Heading from './Heading';
+import TouristInfoCards from './TouristInfoCards';
+import Bookings from './Bookings';
+import Restaurant from './Restaurant';
+import Footer from './Footer';
+import './App.css';
+import './grid.css';
```
40thieves: It's a little thing but I like the spacing & organisation here 👍

src/App.js
```diff
@@ -1,16 +1,45 @@
...
 
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">CYF Hotel</header>
+        <Heading />
         <Bookings />
+       <div className='row'>
+        <TouristInfoCards
```
40thieves: `TouristInfoCards` (with an "s" at the end) implies that this component shows multiple cards, but in fact it just shows 1.

I would either rename it to be singular or (even better I think) rename them to be singular and also extract out showing all three cards to another component named `TouristInfoCards`

src/Button.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+
+function Button() {
+
+    return(
+        <div className="search-row">
+            <input
```
40thieves: This isn't quite right I think - the component is named "Button" but it also includes the input box and the wrapping div.

src/Search.js
```diff
@@ -13,11 +14,10 @@ const Search = () => {
             <div className="search-row">
               <input
                 type="text"
-                id="customerName"
                 className="form-control"
-                placeholder="Customer Id"
+                placeholder="Coustomer ID"
               />
-              <button className="btn btn-primary">Search IDs</button>
+              <Button content="search IDs" />
```
40thieves: Here we're passing a `content` prop but it's not being used in the component at all. (You can test this by changing it to be `content="This doesn't show!"`).

How could we fix that?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+const TouristInfoCards = props => {
+  return (
+    <div className="xs-col-12 sm-col-4">
+      <div class="card" style={{ width: '18rem' }}>
```
40thieves: You've got the one above correct, but remember that you have to use `className` instead of `class` when using React!

src/Orders.js
```diff
@@ -0,0 +1,32 @@
...
+    super(props);
+    this.state = {
+      order: 0
+    };
+  }
+
+  addOrder = () => {
+    this.setState(previousState => {
+      return {
+        order: previousState.order + 1
```
40thieves: Great! Well done :)

src/Restaurant.js
```diff
@@ -1,17 +1,19 @@
...
-};
+class Restaurant extends Component {
+  render() {
+    const pizzas = 0;
+    return (
+      <div>
+        <h3>Restaurant Orders</h3>
+        <ul>
+          <Orders orderType="Pizzas" />
+          <Orders orderType="Salads" />
```
40thieves: Can you see how making our component reusable (but configurable via props) is really useful?

src/SearchResults.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  toggle = e => {
+    
+    console.log('click')
+    let line = e.target.parentNode.id;
+    this.setState(previousState => ({
+      lineId: line
+    }));
+  e.target.parentNode.style.backgroundColor = 'lightblue';
```
40thieves: This is something that you have to be careful about in React: you're directly manipulating the DOM. This is potentially a problem because now React doesn't know that your change is deliberate and will wipe away your change next time your render (anywhere in the app). Your manual DOM change is in the "real DOM", but not in React's virtual DOM and so it removes your manual change.

PR #59, opened 2019-04-13
najla-hotel-project

NajlaSoltana:
job done!!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,16 +1,40 @@
 import React, { Component } from "react";
 
+
 import Bookings from "./Bookings";
 import Restaurant from "./Restaurant";
+import Heading from "./Heading";
+import TouristInfoCards from "./TouristInfoCards";
+import Footer from "./Footer";
 import "./App.css";
 
 class App extends Component {
   render() {
     return (
       <div className="App">
+        <Heading />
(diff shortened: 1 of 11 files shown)


src/App.js
```diff
@@ -1,16 +1,40 @@
...
 
 class App extends Component {
   render() {
     return (
       <div className="App">
+        <Heading />
         <header className="App-header">CYF Hotel</header>
+        
         <Bookings />
+        <TouristInfoCards
```

src/Button.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function Button() {
+  return  <button className="btn btn-primary">Search IDs</button>
```
40thieves: How could we make this component more reusable? Right now it will always show "Search IDs" as the button text. Which isn't very useful if we wanted to use it somewhere else in the app. How could we change this so that it can be reused in multiple different places?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+    return (
+        <div class="card" style={{ width: "18rem" }}>
```
40thieves: Remember that when writing a React application you have to convert `class` to `className`

PR #60, opened 2019-04-13
nina-hotel-project

Ninaa117:
:)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,16 +1,47 @@
 import React, { Component } from "react";
-
+import Heading from "./Heading";
 import Bookings from "./Bookings";
 import Restaurant from "./Restaurant";
+import TouristInfoCards from "./TouristInfoCards";
+import SearchResults from "./SearchResults";
+import Footer from "./Footer";
 import "./App.css";
 
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">CYF Hotel</header>
+        <Heading />
         <Bookings />
+        <TouristInfoCards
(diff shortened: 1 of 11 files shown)


src/App.js
```diff
@@ -1,16 +1,47 @@
...
 import "./App.css";
 
 class App extends Component {
   render() {
     return (
       <div className="App">
-        <header className="App-header">CYF Hotel</header>
+        <Heading />
         <Bookings />
+        <TouristInfoCards
```

src/Bookings.js
```diff
@@ -13,7 +13,8 @@ export default class Bookings extends Component {
       <div className="App-content">
         <div className="container">
           <Search search={this.search} />
-          {/* <SearchResults results={FakeBookings} /> */}
+          <SearchResults results={FakeBookings} />
+          {/* <SearchResults /> */}
```
40thieves: This old commented out `<SearchResults>` can be removed now :)

src/Button.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const Button = props => {
+  return <button className="btn btn-primary">{props.caption}</button>;
```
40thieves: Nice work making this component reusable!

src/TouristInfoCards.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div class="card" style={{ width: "18rem" }}>
```
40thieves: Remember that when writing a React application you have to convert `class` to `className`!

PR #72, opened 2019-05-23
Update lesson 1

llh1:
- Improve instructions for Lesson 1
- Change class components into functional components (class components haven't been introduced yet after class 1)


NadaKartouch:
looks good thanks Loic

PR #74, opened 2019-06-09
Update instructions for Lesson 3: replace refs with controlled component for the search input

llh1:


NadaKartouch:
Looks good to me

PR #75, opened 2019-06-15
completed Hotel exercise

youssef910:
completed Hotel exercise


src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,37 @@
...
+            <a href={city.Link} className="btn btn-primary">
+              Go {city.CityName}
+            </a>
+          </div>
+        </div>
+      ))}
+    </div>
+  );
+};
+export { TouristCardsLinksAndImages, TouristInfoCards };
```
llh1: One improvement here, you could extract `TouristCardsLinksAndImages` variable into a different file. It doesn't need to be in this file as you pass it as a props to `TouristInfoCards`. So in this file, you just need the definition of the component `TouristInfoCards`. It would make this file clearer and more generic.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+  {
+    CityName: "London",
+    Link: "https://www.visitlondon.com/",
+    Photo: "https://cdn.londonandpartners.com/fl/assets/33374-310x174.jpg"
+  }
+];
+const TouristInfoCards = props => {
+  return (
+    <div className="Card-Container">
```
llh1: Try to follow one unique naming convention in your code. So here for the className value, as you use snakecase format in the rest of the file, you could rename it `card-container` instead of `Card-Container` for consistency.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+var TouristCardsLinksAndImages = [
+  {
+    CityName: "Glasgow",
+    Link: "https://peoplemakeglasgow.com/",
+    Photo: "https://peoplemakeglasgow.com/templates/people/images/logo.png"
+  },
```
llh1: The usual naming convention for Javascript keys is to use camelcase but with the first letter being in lower case. Not a big deal, but it's more about being consistent with the JS convention: so `cityName` instead of `CityName` etc...

src/components/Bookings.js
```diff
@@ -0,0 +1,78 @@
...
+    if (StartTyping === true) this.setState({ searching: true });
+  };
+  // to push the new array transfered from input file
+  addBooking = booking => {
+    const updatedBookings = this.state.bookings;
+    updatedBookings.push(booking);
+    this.setState({ bookings: updatedBookings });
+  };
+  render() {
+    if (this.state.isLoading === true) return <span>LOADING.....</span>;
```
llh1: Improvement: you can remove the `=== true`, it will evaluate already the expression as true or false in your if statement.

src/components/SearchResults.js
```diff
@@ -0,0 +1,124 @@
...
+              ))}
+          </tbody>
+        ) : (
+          <tbody>
+            {this.props.fakeBookingsList
+              .sort(this.compareFunction)
+              .reverse()
+              .map((item, index) => (
+                <TableRow key={index} item={item} />
+              ))}
```
llh1: The above code is almost the same duplicated 2 times. How to remove duplication? The only difference is the user of the `reverse()` function. Can you extract some part in a function to avoid duplication in the render() function?

src/components/SearchResults.js
```diff
@@ -0,0 +1,124 @@
...
+    super(props);
+    this.state = {
+      isClicked: false,
+      style: {}
+    };
+  }
+
+  onClicked = () => {
+    this.setState(() => {
+      if (this.state.isClicked === false)
```
llh1: This is correct. However, another way would be to use a `previousState` parameter in the function you use in `setState`. So you can write `previousState.isClicked` instead of `this.state.previousState`. It's generally considered a better practice.

src/components/SearchResults.js
```diff
@@ -0,0 +1,124 @@
...
+      sortAscending: true
+    };
+  }
+  onClicking = () => {
+    this.setState({ style: { backgroundColor: "skyblue" } });
+  };
+  compareFunction = (a, b) => {
+    if (a[this.state.sortType] > b[this.state.sortType]) return 1;
+    if (a[this.state.sortType] < b[this.state.sortType]) return -1;
+    return 0;
```
llh1: A shorter way to write this: 
```
compareFunction = (a, b) => {
   return a[this.state.sortType] > b[this.state.sortType] ? 1 : -1;
}
```

src/components/SearchResults.js
```diff
@@ -0,0 +1,124 @@
...
+  };
+  handleClickToSort = event => {
+    let text = event.target.textContent;
+    text = text.charAt(0).toLowerCase() + text.slice(1);
+    this.setState({
+      sortType: text,
+      changeArowDirection: !this.state.changeArowDirection,
```
llh1: Do you need this state? I cannot find where it is used.

src/components/Inputs.js
```diff
@@ -0,0 +1,97 @@
+import React from "react";
+// to create an input form to insert new guest
+class Inputs extends React.Component {
```
llh1: Try to find a more specific name for this component. `Inputs` tells me that it's going to be several inputs but nothing more. Here, you actually have a form to add a new booking for a guest, how about naming this component `AddBookingGuestForm` ?

src/components/Inputs.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+  // creating new object of entries
+  handleChange = event => {
+    const updateduser = this.state.user;
+    updateduser[event.target.name] = event.target.value;
+    console.log(this.state.user);
+    this.setState({ user: updateduser });
+    // this.setState({
+    //   user: { ...this.state.user, [event.target.name]: event.target.value }
+    // });
```
llh1: Keep your code clean, if some code is unused, remove it instead of commenting it.

src/components/Inputs.js
```diff
@@ -0,0 +1,97 @@
...
+        checkInDate: "",
+        checkOutDate: ""
+      }
+    };
+  }
+  // creating new object of entries
+  handleChange = event => {
+    const updateduser = this.state.user;
+    updateduser[event.target.name] = event.target.value;
+    console.log(this.state.user);
```
llh1: Remove console.log from your code, this is used only for debugging purposes.

src/components/Inputs.js
```diff
@@ -0,0 +1,97 @@
...
+  // on submit add the value of the user object to the function addBoking "transfered as a props from bookings file"
+  handelSubmit = e => {
+    e.preventDefault();
+    this.props.addBooking(this.state.user);
+  };
+  render() {
+    return (
+      <div>
+        <form
+          className="App_input_lest"
```
llh1: Adopt one same naming convention for your className across your code.

src/components/Search.js
```diff
@@ -0,0 +1,59 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      searchInput: "",
+      SearchingActive: false
+    };
+  }
+
+  handleSearcInput = event => {
+    this.props.searching(this.state.SearchingActive);
```
llh1: JS naming convention: should be `searchingActive` instead of `SearchingActive`

src/components/Bookings.js
```diff
@@ -0,0 +1,78 @@
...
+          String(item.surname).toLowerCase() ===
+            String(searchVal)
+              .trim()
+              .toLowerCase()
+        );
+      })
+    });
+  };
+  //to make the table displayed in case of not searching and disappear in case searching and then display the result on submit
+  searching = StartTyping => {
```
llh1: JS naming convention: should be `startTyping` instead of `StartTyping`

src/components/Bookings.js
```diff
@@ -0,0 +1,78 @@
...
+      return (
+        <div className="App-content">
+          <div className="container">
+            <Search search={this.search} searching={this.searching} />
+            {/* to make the table displayed in case of not searching and disappear in case searching and then display the result on submit */}
+            {this.state.searching ? (
+              <SearcResults fakeBookingsList={this.state.filteredarray} />
+            ) : (
+              <SearcResults fakeBookingsList={this.state.bookings} />
+            )}
```
llh1: You will have a problem here when the user starts to search something, it removes all the table. I think the table shouldn't change until the user clicks on the button Search IDs, otherwise this button is not useful and the user experience is a bit confusing. Then even if I delete all the search and click on the button Search IDs, i have no ways to get the original results back on the screen. Can you solve this :) ?

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+var TouristCardsLinksAndImages = [
```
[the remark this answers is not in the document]
youssef910: i did it in the first week and at this week you were didn't introduce the const to us and i forgot to edit it but after that i am using  const

src/components/TouristInfoCards.js
youssef910: edited it

src/components/TouristInfoCards.js
youssef910: sometimes the uppercase letters is added unconsciously :D

src/components/TouristInfoCards.js
youssef910: uppercase again :D

src/components/Bookings.js
youssef910: done

src/components/SearchResults.js
youssef910: done

src/components/SearchResults.js
youssef910: no it was an unsuccessful  try to add  up and down arrow   changing with the sorting

PR #89, opened 2020-05-03
Updating homework for React Hooks

llh1:
**Objectives:** update the instructions + initial code base of cyf-hotel-react for React Hooks.

Progress:
- Few tweaks in Lesson 1.
- Lesson 2: updated for React Hooks.
- Lesson 3: updated for React Hooks.


README.md
```diff
@@ -32,47 +32,52 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+8. Within `src/App.js`, render the `<Restaurant />` component (that is provided for you in `src/Restaurant.js`) underneath the `<Bookings />` component.
 
-9. Add a method to the `Restaurant` class named `addOrder`. Use `console.log` to log a "Add order" message. Remember to use the `addOrder = () => {}` syntax. Add a `onClick` handler to the Add `<button>` that calls `this.addOrder`. Ensure that clicking on the button logs your "Add order" message in the console.
+9. At the moment, the number of pizzas a guest can order is static and set to 0, even if they click on the 'Add' button. We will change that in the following to let a guest add more pizzas to their order. First, declare a new state variable `orders` along the function to set the orders state `setOrders`. The initial value of the `orders` state should be **0**. Use the new `orders` variable instead of the `pizzas` variable (that you can now delete) and verify the number of ordered pizzas it still **0** on the screen. Hint: you need to use the React function `useState` to create a state variable. Remember to import the function at the top with `import React, {useState} from "react";`.
```
zubhav, line 43:
```suggestion
9. At the moment, the number of pizzas a guest can order is static and set to 0, even if they click on the 'Add' button. We will change that in the following to let a guest add more pizzas to their order. First, declare a new state variable `orders` along with the function to set the orders state `setOrders`. The initial value of the `orders` state should be **0**. Use the new `orders` variable instead of the `pizzas` variable (that you can now delete) and verify the number of ordered pizzas it still **0** on the screen. Hint: you need to use the React function `useState` to create a state variable. Remember to import the function at the top with `import React, {useState} from "react";`.
```

zubhav: Would it be worth adding a link to docs for useState? Or will that be too much of a hint? https://reactjs.org/docs/hooks-state.html

zubhav, line 43: orders along *with* the function

zubhav: Is it worth adding a link to the useState docs here? Or is that too much of a hint? https://reactjs.org/docs/hooks-state.html

README.md
```diff
@@ -32,47 +32,52 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+15. Within the `<SearchResults />` component or its child components, add an `onClick` handler to each row in the table (hint: on the `<tr>` element). When clicked, the row is "selected" and highlighted with a different colour. When clicked again, the row is unselected and the coloured highlighting is removed. Hint: use a new state variable for each row to record if the row is selected or not, and use this value to set a class to the `className` prop of the row.
 
 ### Lesson 3
 
 15. Within your `<Header />` component, render the `<Clock />` component (that is provided for you in `src/Clock.js`). Fix the problem where the `setTimeout` timer is not **cleared** if the component is **unmounted**. Hint: look at the Clock exercise you did in class.
+    **TODO: is this question still relevant with React Hooks? Need to verify this is still in line with what's taught in the syllabus.**
```
zubhav, line 60: I believe this is still relevant with hooks. You would clear the timeout within the clean up of a useEffect. https://reactjs.org/docs/hooks-effect.html "Effects with Cleanup" section

zubhav: Not sure if this will be too complicated in the scope of the lesson. Lets discuss this..

README.md
```diff
@@ -32,47 +32,52 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+16. In the following, we will implement the functionality to search for a customer name given the input typed in the customer name field. In the `src/Search.js` file, declare a new state variable named `searchInput` with the corresponding setter function `setSearchInput` (hint: use the React function `useState`). The initial value of the `searchInput` variable can be an empty string. Add a `value` property to the `<input>` tag that is set to the new `searchInput` state variable. Create a new function `handleSearchInput` taking an `event` parameter. This function should use the `setSearchInput` function to update the state variable `searchInput` with what the user typed in the input field (hint: use `event.target.value` to get the input value). Finally, add a `onChange` prop to the `<input>` tag that is set to the function `handleSearchInput`.
```
zubhav, line 62: given the input typed in the customer name field
==>
given the text that has been typed into the customer name field

README.md
```diff
@@ -32,47 +32,52 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
 
-19. Still in the `<Bookings />` component, implement the `search` method. It must use the `searchVal` (that you just passed from the `<Search />` component) to **filter** the search results. The filter function should return results where `firstName` or `surname` match `searchVal`. Once filtered, use `this.setState` to update the results rendered in `<SearchResults />`.
+20. Now instead of getting the existing bookings from the file `data/fakeBookings.json`, we will get the bookings from a remote API. In the `<Bookings />` component, use the React function `useEffect` to console log some text only when the page first renders on the screen. Verify that when you refresh the page, the text appears once in the console. Then, in the `useEffect` function, use the `fetch()` function to get data from `https://cyf-react.glitch.me`. Hints:
```
zubhav, line 70: console log
==>
`console.log()`

README.md
```diff
@@ -32,47 +32,52 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+21. Now show a _loading state_ in `<Bookings />` while the data from the server is being fetched. To test this, try loading data from `https://cyf-react.glitch.me/delayed`, which has a 5 second delay before returning the data. You will need to use another state to record when your application is loading data (this can be a boolean) and display a loading message whenever the application is loading data. Hint: try looking at your Pokemon app that you worked on in class for an example.
 
-22. Finally, display an error message in `<Bookings />` if there is an HTTP error when fetching data from the server. To test this, try loading data from `https://cyf-react.glitch.me/error`, which will return a 500 HTTP error. Hint: Try looking at your Pokemon app that you worked on in class for an example
+22. Finally, display an error message in `<Bookings />` if there is an HTTP error when fetching data from the server. To test this, try loading data from `https://cyf-react.glitch.me/error`, which will return a 500 HTTP error. Hint: Try looking at your Pokemon app that you worked on in class for an example.
```
zubhav, line 80: Do we need to provide information for `.catch()` when the fetch fails or should students be comfortable with this?

README.md
40thieves: I have indeed added it to my PR: https://github.com/CodeYourFuture/syllabus/pull/417, but with an open question about whether it's worth including.

README.md
40thieves: I would have to double check with what was taught in JS2/3 but I believe they would be familiar with `.catch()`. Given that this is the last exercise, I think leaving off some hints is likely fine.

README.md
llh1: At this stage, they would have seen and practice useState already during the class, so all good i think.

README.md
40thieves: Just noting that we decided to remove unmounting/cleaning up from the main syllabus and adding it as extra material. So I think these exercises should now be removed.

40thieves:
I like the new formatting! Thanks @llh1 🙂

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+**Test:** Verify the number of ordered pizzas it still **0** on the screen.
+
+#### 10. Add more pizzas
+
+**Instructions:** In the `<Restaurant />` component, create a new function named `orderOne`. The `orderOne` function doesn't take any parameters and should use the `setOrders` function to increment the `orders` state variable by 1. Then, add a `onClick` handler to the Add `<button>` that calls the `orderOne` function when it's being clicked.
+
+**Test:** Try to click on the Add button a few times and verify that the number of pizzas increases accordingly.
+
+#### 11. Extract the Add button in its own component
```
40thieves, line 99:
```suggestion
#### 11. Extract the Add button to its own component
```

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+**Instructions:** Extract the `<button>` currently in the `<Restaurant />` component to a new component named `RestaurantButton`. Pass the `orderOne` function as a prop to the `<RestaurantButton />` component and use this prop in the `onClick` handler.
+
+**Test:** Clicking the button should still increment the number of pizzas.
+
+#### 12. Extract pizza order in its own Order component
```
40thieves, line 105:
```suggestion
#### 12. Extract pizza order to its own Order component
```

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+**Test:** Make sure the pizza order is still rendered on the page and that clicking on the "Add" button still increments the number of orders.
+
+#### 13. Render more orders
+
+**Instructions:** Pass a new prop named `orderType` to the `<Order />` component with the value "Pizzas". Then render the `orderType` prop instead of "Pizzas" in the `<Order />` component. Make sure that "Pizzas" is still displayed on the screen. In the `<ul>` list of the `<Restaurant />` component, render 2 others `<Order />` components but this time pass different values for the `orderType` prop: "Salads" and "Chocolate cake".
+
+**Test:** For each order, the number of items can be incremented independently. Verify that you are able to explain what is happening.
+
+#### 14. Holding passing bookings from a state variable
```
40thieves, line 117: I think this is supposed to be:

```suggestion
#### 14. Passing bookings from a state variable
```

?

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+
+**Hints:**
+
+- Replace `FakeBookings` in the bookings state and initialise it with `[]` (because we haven't fetched any results yet!)
+- After calling the `fetch()` function, use `.then()` to handle the response. Try looking at your Pokemon app that you worked on in class for an example
+- When the response comes back, use `setBookings` to update the results
+
+**Test:** Verify the customers data are still displayed correctly in the table.
+
+#### 16. Highlight booking row when clicked
```
40thieves, line 137: Historically, this has been the hardest exercise (I think most students haven't been able to complete it). So I think I would suggest making it a stretch goal.

On the other hand I think it's also the exercise which *teaches* the most: it really get students to "think in components", because the simplest solution is to push the selected state down to individual rows in the table. Most students approach it by trying to model multiple selection states in the parent.

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
 
-5. Create a `<SearchResults />` component that shows hotel bookings in a `<table>` element. Each booking will have an id, title, first name, surname, email, room id, check in date and check out date. You can make up data to show in the table. Then show `<SearchResults />` component within the `<Bookings />` component that is provided for you. Be sure to split out your components into small well-named components, similar to the method used in exercise 1. Hint: You will find some useful `<table>` examples in the [Bootstrap documentation for tables](https://getbootstrap.com/docs/4.2/content/tables/#examples)
+**Test:** Verify that when you enter an existing first name or surname and submit the form, the results are filtered accordingly in the customers table.
 
-6. Instead of using your hard-coded data in the `<SearchResults />` component, load data from the `src/data/fakeBookings.json` file in the `<Bookings />` component and pass it as a prop to `<SearchResults />`. Hint: look in the `<Bookings />` component for how to import data from a JSON file.
+#### 20. Highlight only one row when selected
```
40thieves, line 169: As discussed above, the best solution to task 16 is to push selection state down to the children. Moving it back up again makes the task easier in some ways.

Reading further, I see why you have done this here though: because it fits well into task 21.

Perhaps we could just add a button to the table row to show in the `CustomerProfile`? Or we could just leave it like this, I don't necessarily think it's a problem per-se.

package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "react-scripts": "3.0.1"
+    "react-scripts": "^3.4.1"
```
40thieves, line 8: Thanks for doing this 👍

README.md
```diff
@@ -28,54 +40,174 @@ A hotel booking application in React. Homework for the [CodeYourFuture React mod
...
+#### 21. Display a customer profile
 
-9. Add a method to the `Restaurant` class named `addOrder`. Use `console.log` to log a "Add order" message. Remember to use the `addOrder = () => {}` syntax. Add a `onClick` handler to the Add `<button>` that calls `this.addOrder`. Ensure that clicking on the button logs your "Add order" message in the console.
+**Instructions:** Create a new `<CustomerProfile />` component. This component should be rendered next to the table in the `<SearchResults />` component. This component can be passed one prop `id`, corresponding to the user id which is selected (as defined in the task 20). Initially, the `<CustomerProfile />` component doesn't show anything. When a customer is selected in the table, fetch the corresponding profile from `https://cyf-react.glitch.me/customers/<ID>` in the `<CustomerProfile />` component. A customer profile should show the customer ID, their email, their nationality, if they are VIP and their phone number.
```
40thieves, line 177: We picked this up on Slack, but this will require changes to the [Glitch app](https://glitch.com/~cyf-react) that powers the little backend. I've invited @llh1 to edit that project.

llh1: I've added the /customers/:id endpoint to the Glitch project, thanks 👍

README.md
llh1: Cool i like the suggestion to add a button, it may make things a bit easier as they won't have to change their highlight implementation from task 16.

README.md
llh1: I also think it's an interesting exercise, it appears as the last task of the homework for lesson 2, that should be okay i think? I'm going to introduce a button as you suggested for the task 21, so this task 16 is not a blocker if students don't get it 100% right.

40thieves: Sounds good, thanks.

The other advantage of it being in week 2, is that we can explain the solution at the beginning of the week 3 lesson.

README.md
llh1: I've updated task 20 and 21, let me know what you think.

PR #92, opened 2020-05-20
Meisam

Erwin6997:


40thieves:
Hi @Erwin6997! Is your homework now complete? If so, could you mark this PR as "Ready for review"

There's a button below to do that:

Screen Shot 2020-05-25 at 17 00 43

src/App.js
```diff
@@ -1,13 +1,26 @@
...
+import Footer from "./Footer";
+import Restaurant from "./Restaurant";
+
 import "./App.css";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <heading className="App-header">
```
40thieves, line 12: Exercise 2 asks you to "extract" the `<header>` to a component. That means you should move it to a separate component (in a separate file). Do you think you can do that?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+const TouristInfoCards = () => {
```
40thieves, line 3: This component is responsible for rendering all of the tourist info cards.

Perhaps we have a component that is responsible for rendering just a single card? What props would you pass to this component?

src/Footer.js
```diff
@@ -0,0 +1,19 @@
...
+  "hello@fakehotel.com",
+  "0123 456789"
+];
+const Footer = () => {
+  return (
+    <footer className="footer">
+      <ul>
+        {footer.map((foot, index) => (
+          <li key={index}>{foot}</li>
+        ))}
```
40thieves, line 14: Interesting solution here too! 🙂

src/ShowAll.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import FakeBookings from "./data/fakeBookings.json";
```
40thieves, line 2: If you look carefully at the instructions for Exercise 6, you'll see it says:

> load data from the `src/data/fakeBookings.json` file in the `<Bookings />` component and pass it as a prop to `<SearchResults />`

Here we're loading the json directly inside the component and not passing it as a prop. Could you fix this so that the data is passed from `Bookings` to this component?

40thieves, line 2: If you look carefully at the instructions for Exercise 6, you'll see it says:

> load data from the `src/data/fakeBookings.json` file in the `<Bookings />` component and pass it as a prop to `<SearchResults />`

Here we're loading the json directly inside the component and not passing it as a prop. Could you fix this so that the data is passed from `Bookings` to this component?

src/App.js
Erwin6997: ok yes. thanks. I did now.

src/Nights.js
```diff
@@ -0,0 +1,10 @@
+import react, { timeline, timeout } from "react";
+import moment from "moment";
+
+const Nights = props => {
+  let to_data = moment(props.timeout);
+  let form_date = moment(props.timeline);
+  let diff_days = to_data.diff(form_date, "days");
+  return diff_days;
+};
```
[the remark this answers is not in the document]
Erwin6997: it's a very short way az I think.
I used this component again in the result and maybe I need this more for calculating the bill.

src/TouristInfoCards.js
Erwin6997: Nothing , Do you mean I should make one card and props transfer the data which is about three cards in the component?

src/ShowAll.js
Erwin6997: I did. I passed the Data to the two components.

Erwin6997: The same I did

src/Order.js
```diff
@@ -0,0 +1,26 @@
...
+
+const Order = props => {
+  const [pizzas, setPizzas] = useState(0);
+  const [orders, setOrders] = useState(0);
+  const [price, setPrice] = useState(0);
+  const [bill, setBill] = useState(price);
+
+  const orderOne = () => {
+    setOrders(orders + 1);
+    setPrice(props.price + price);
```
40thieves, line 12: Nice work, I like the addition of `price` state 🙂

src/Bookings.js
```diff
@@ -1,17 +1,24 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import SearchButton from "./Search.js";
```
40thieves, line 2: Could you rename this import to just `Search`, otherwise it's a bit confusing.

src/Search.js
```diff
@@ -1,11 +1,10 @@
 import React, { useState } from "react";
 import SearchResults from "./SearchResults";
-const SearchButton = () => {
+const SearchButton = props => {
```
40thieves, line 3: Does this component actually represent a `SearchButton`? What do you think it should be renamed to?

src/ShowAll.js
```diff
@@ -1,41 +1,44 @@
-import React from "react";
-import FakeBookings from "./data/fakeBookings.json";
+import React, { useState } from "react";
 import Nights from "./Nights";
-const ShowAll = () => {
+const ShowAll = props => {
+  const [clicked, setClicked] = useState(false);
```
40thieves, line 4: This is a good attempt, but it looks like if you click on any row, then all of them become highlighted.

We'll look at a simple solution on Sunday, but the idea is to think carefully about how we could split up our components so that they can have their own state.

src/Restaurant.js
```diff
@@ -1,15 +1,31 @@
...
+            <th> Number</th>
+            <th>All price</th>
+            <th> Add</th>
+          </tr>
+
+          <Order orderType={"Pizzas"} price={9.99} />
+
+          <Order orderType={"Salads"} price={5.99} />
+
+          <Order orderType={"Chocolate cake"} price={3.99} />
```
40thieves, line 22: Nice work! 🍰

src/Order.js
Erwin6997: thanks. I was trying to make Bill but I couldn't addition of all the price.

src/Bookings.js
Erwin6997: Done

src/Search.js
Erwin6997: Search is better. 
 ;)

src/ShowAll.js
Erwin6997: Yes. I got this problem when I selected.

PR #93, opened 2020-05-20
Nader-hotel-react-project

naderakhgari:


src/App.css
```diff
@@ -39,19 +52,57 @@
...
+  align-items: center;
+  background-color: #222;
+  height: 50px;
+  padding: 20px;
+  color: white;
+  font-family: Arial, Helvetica, sans-serif;
+  font-size: 1em;
+  font-weight: bold;
+  bottom: 0;
+  margin-bottom: 0;
```
zubhav, line 99: Is the `bottom: 0;` line needed here? Also, what about `margin-bottom: 0;`? Try to remove each one individually and see what happens!

src/App.js
```diff
@@ -1,15 +1,23 @@
...
+import Restaurant from "./Restaurant";
+import Footer from "./Footer";
+import TouristInfoCards from "./TouristInfoCards";
+import Heading from "./Heading";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
```
zubhav, line 12: Good structure!

src/App.js
```diff
@@ -1,15 +1,23 @@
...
       <Bookings />
+      <Restaurant />
+      <Footer />
     </div>
   );
 };
 
 export default App;
+
+// key={props.key}
```
zubhav, line 23: We can remove this temporary comment :)

src/Bookings.js
```diff
@@ -1,18 +1,17 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
```
zubhav, line 4: Since this isn't a component, it is good convention to use camelcase naming for this. So `fakeBookings` would be more suitable. 

Try to reserve capital casing for React components.

This is a minor very minor comment 👍

src/Footer.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+const Footer = () => {
+  let footerEls = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
+  return (
+    <ul className="footer  footer-list">
```
zubhav, line 9: Very interesting use of `map` here to construct your footer. Good job!

src/Search.js
```diff
@@ -26,4 +26,8 @@ const Search = () => {
   );
 };
 
+const SearchButton = () => {
```
zubhav, line 29: It's a good idea to extract this component into its own file.

Also I'd like you to have a think about the following:

- What if we made this component even more generic so that we could pass in any text in props?
- The component would have manage styling and the text for the component would come from a parent component
- Would we still have to call this `SearchButton` then?
- Think about how we could we implement this and what benefits this has!

src/SearchResults.js
```diff
@@ -0,0 +1,47 @@
...
+          <td>room id</td>
+          <td>check in date</td>
+          <td>check out date</td>
+          <td>nights</td>
+        </tr>
+      </thead>
+      <tbody className="table-body table-striped">
+        {props.results.map(result => {
+          return (
+            <tr key={result.id}>
```
zubhav, line 23: 👍

src/SearchResults.js
```diff
@@ -0,0 +1,47 @@
...
+              <td>{getDuration(result.checkOutDate, result.checkInDate)}</td>
+            </tr>
+          );
+        })}
+      </tbody>
+    </table>
+  );
+};
+
+function getDuration(start, end) {
```
zubhav, line 41: 🎉

src/TouristInfoCards.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+
+const TouristInfoCards = () => {
```
zubhav, line 3: This component is responsible for rendering all the Tourist Info cards. Could we have a component that is responsible for rendering a single card?

src/App.css
naderakhgari: I think you're right bottom=0 is not needed but `  margin-bottom: 0;` needed to remove the bottom margin under footer

src/App.js
naderakhgari: it's right, done

src/Bookings.js
naderakhgari: Yes it's a good point

src/Search.js
naderakhgari: the first part I understood and done but the last four point that you mentioned did not understand.

src/TouristInfoCards.js
naderakhgari: Yes it's a good idea. I made a new component called Card and passed the city as a prop for it

src/Bookings.js
```diff
@@ -1,17 +1,24 @@
...
+import "./App.css";
 
 const Bookings = () => {
+  useEffect(() => {
+    console.log("Hi, this is the first using useEfct");
+    fetch("https://cyf-react.glitch.me")
+      .then(res => res.json())
+      .then(data => setBookings(data));
+  }, []);
+  const [bookings, setBookings] = useState([]);
```
40thieves, line 13: Nice work! There's one little thing that I would suggest changing to follow the *convention* of React Hooks. We normally put all of our `useState` calls at the top of the component:

```js

  useEffect(() => {
    console.log("Hi, this is the first using useEfct");
    fetch("https://cyf-react.glitch.me")
      .then(res => res.json())
      .then(data => setBookings(data));
  }, []);
```

src/Card.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Card = props => {
```
40thieves, line 3: 👍

src/Order.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+import "./App.css";
+
+const Order = props => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
+    setOrders(orders + 1);
```
40thieves, line 8: Excellent, well done!

src/Restaurant.js
```diff
@@ -1,15 +1,16 @@
...
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
-      </ul>
+      <div className="orders">
+        <Order orderType="Pizzas" />
+        <Order orderType="Salads" />
+        <Order orderType="Chocolate cake" />
```
40thieves, line 12: 🍰

src/Row.js
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from "react";
+
+const Row = props => {
+  return (
+    <tr
+      onClick={() => props.setSelectedRow(props.row.id)}
+      className={props.isSelected ? "selected-row" : "unselected-row"}
```
40thieves, line 7: This is **really** close to the correct solution, but not quite! The exercise asks you to let rows be highlighted *independently*. This means that  multiple rows can highlighted at once.

Your solution doesn't do that, but with some small changes it would. Can you figure out how to change it?

PR #95, opened 2020-05-21
Ellie-Hotel-React

ellietms:
This is my pull request for first react project!

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from "react";
+import Heading from "./Heading";
 
 import Bookings from "./Bookings";
 import "./App.css";
@@ -6,7 +7,7 @@ import "./App.css";
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
       <Bookings />
     </div>
   );
--- a/src/App.js
+++ b/src/App.js
@@ -2,14 +2,21 @@ import React from "react";
 import Heading from "./Heading";
 import TouristInfoCards from "./TouristInfoCards";
 import Bookings from "./Bookings";
+import Footer from "./Footer";
 import "./App.css";
 
 const App = () => {
+  const addresses = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
   return (
     <div className="App">
       <Heading />
       <TouristInfoCards />
       <Bookings />
+      <Footer keys={addresses} />
     </div>
   );
 };
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,6 @@
 import React from "react";
 import Heading from "./Heading";
-
+import TouristInfoCards from "./TouristInfoCards";
 import Bookings from "./Bookings";
 import "./App.css";
 
@@ -8,6 +8,7 @@ const App = () => {
   return (
     <div className="App">
       <Heading />
+      <TouristInfoCards />
       <Bookings />
     </div>
   );
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -9,8 +9,8 @@ const Bookings = () => {
   };
 
   return (
-    <div className="App-content">
-      <div className="container col-xs-12 col-sm-12  px-auto col-md-12 col-lg-10 mx-auto">
+    <div>
+      <div className="container col-sm-12">
         <Search search={search} />
         <SearchResults results={FakeBookings} />
       </div>
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,7 +1,8 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import { Table } from "reactstrap";
+import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
   const search = searchVal => {
@@ -12,7 +13,7 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
       </div>
     </div>
   );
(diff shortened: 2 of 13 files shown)


src/App.js
```diff
@@ -1,13 +1,22 @@
...
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer keys={addresses} />
```
illicitonion, line 19: Keys is not the most descriptive name for what this prop does - can you think of a name which better describes what the values are / how they'll be used?

src/SearchResults.js
```diff
@@ -0,0 +1,48 @@
...
+          <th className="mx-auto bg-dark text-white pl-2">Check in date</th>
+          <th className="mx-auto bg-danger text-white pl-2">Check out date</th>
+          <th className="mx-auto bg-dark text-white pl-2">Nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map((element, index) => {
+          return (
+            <tr key={index} className="col-12">
+              <th scope="row" className="bg-dark text-white">
```
illicitonion, line 24: Completely optional idea to make this even tidier: Can you make it so that you don't need to manually alternate the `bg-danger` and `bg-dark` classes? (And so that you don't need to put `text-white` on each cell)?

Hint: Look up the `first-child` and `nth-child` selectors

src/TouristInfoCards.js
```diff
@@ -0,0 +1,94 @@
...
+  CardText,
+  CardBody,
+  CardTitle,
+  Button
+} from "reactstrap";
+
+const TouristInfoCards = props => {
+  return (
+    <div className="col-12 d-lg-flex mx-auto">
+      <Card className="card bg-dark text-white flex-sm-column col-12 ml-auto mr-auto ml-md-4  mr-lg-2 col-md-10 mx-auto mx-md-auto col-lg-4 mt-md-5 mt-4 pr-2">
```
illicitonion, line 14: This is looking great, but you're having to repeat a lot of the same information for all three cards here (e.g. you're saying that they contain a `CardImg` with a `top` property three times). Can you try extracting a component so that you can avoid some of that repetition?

src/SearchResults.js
ellietms: I didn't understand it daniel, how can I fix it?

src/Bookings.js
```diff
@@ -1,18 +1,19 @@
...
-  const search = searchVal => {
-    console.info("TO DO!", searchVal);
-  };
-
+  const [bookings, setBookings] = useState([]);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me/")
+      .then(response => response.json())
+      .then(data => setBookings(data));
+  }, []);
```
40thieves, line 11: Nice work 👍

src/Restaurant.js
```diff
@@ -1,14 +1,12 @@
...
-  const pizzas = 0;
   return (
-    <div>
+    <div className="orders">
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order />
```
40thieves, line 9: This looks good, but exercise 13 asks you to render some additional `Order` components:

> render 2 others <Order /> components but this time pass different values for the orderType prop: "Salads" and "Chocolate cake".

Could you make those changes too?

src/RestaurantButton.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const RestaurantButton = props => {
+  return (
+    <div>
+      <button onClick={props.orderOne} className="btn btn-danger mx-2">
+        Add
+      </button>
+      <button onClick={props.cancelOrder} className="btn btn-warning">
```
40thieves, line 9: I like the Cancel button idea! 🍰

src/SearchResults.js
```diff
@@ -1,44 +1,102 @@
...
+import React, { useState } from "react";
 import { Table } from "reactstrap";
-import moment from "moment";
+import Row from "./row";
 
 const SearchResults = props => {
+  const [active, setActiveClass] = useState(false);
+  function handleClick() {
+    setActiveClass(!active);
+  }
```
40thieves, line 9: Is this state being used? I think it might have been left-over from an old version. If it's not being used, could you remove it?

src/order.js
```diff
@@ -0,0 +1,20 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = () => {
+  const [orders, setOrders] = useState(0);
+  function orderOne() {
+    setOrders(orders + 1);
```
40thieves, line 7: Great work! 🎆

src/row.js
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const Row = ({ element, index }) => {
+  const [active, setActiveClass] = useState(false);
+  function handleClick() {
+    setActiveClass(!active);
+  }
```
40thieves, line 8: This is really excellent work @ellietms! Could you explain why this works?

src/Restaurant.js
ellietms: @40thieves  I didn't see  this, I fixed it

src/SearchResults.js
ellietms: @40thieves  I fixed it, I forgot to add onClick  for it

src/row.js
ellietms: @40thieves  yes, the value of active is false when someone clicks in the rows of the table it is going to change to true and add the classes

40thieves: Yep that's correct 🙂 

Another question for you: why do you think it's necessary to have a `Row` component?

ellietms: @40thieves at first  I Had all of my codes inside of search result component but when I clicked in one row of the table all of the rows changed the colour I really don't know why but when I changed them to another component it worked!

40thieves: Yep, that's exactly the problem. We'll talk more about it on Sunday

ellietms: @40thieves  thank you ali

PR #97, opened 2020-05-23
Abbas CYF Hotel

abbasmokhtari:
This includes the first lesson all 7 segments

### DIFF
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,7 +1,7 @@
 import React from "react";
 import Search from "./Search.js";
 import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
   const search = searchVal => {
@@ -12,8 +12,8 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
-        <SearchResults />
+        <SearchResults results={FakeBookings} />
+        {/*<SearchResults />*/}
       </div>
     </div>
   );
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,6 +1,6 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
+import SearchResults from "./SearchResults.js";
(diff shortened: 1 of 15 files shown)


src/Bookings.js
```diff
@@ -12,7 +12,8 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
+        {/*<SearchResults />*/}
```
illicitonion, line 16: Probably best to remove the commented out code after you're done with it :)

src/SearchButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const SearchButton = () => {
+  return (
+    <div>
```
illicitonion, line 5: Why did you add an extra div around the button? It's not wrong, but I'm also not sure it's necessary, and generally when coding less is better :)

src/Bookings.js
```diff
@@ -12,7 +12,8 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
```
illicitonion, line 15: This passes the `results` prop, but the `SearchResults` component doesn't actually read that prop - instead it separately loads its own copy of `FakeBookings`. Can you change `SearchResults` so that it uses this prop, and doesn't itself import `FakeBookings`?

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import FakeBookings from "./data/fakeBookings.json";
+import moment from "moment";
+
+const SearchResults = () => {
+  return (
+    <div>
```
illicitonion, line 7: Like above, why do you need this `div` tag?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+
+const TouristInfoCards = () => {
+  return (
+    <div className="cards-layout">
+      <div className="card">
```
illicitonion, line 6: It looks like you have three separate copies of the same code to make a card in this component - could you extract a `TouristInfoCard` component to avoid that duplication?

PR #98, opened 2020-05-23
Cyf hotel react/sabia

sabia-m:


src/App.js
```diff
@@ -1,13 +1,25 @@
...
+import TouristInfoCards from "./TouristCards";
+import Footer from "./Footer";
 import "./App.css";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      {/* <header className="App-header">CYF Hotel</header> */}
```
illicitonion, line 13: It looks like you've left quite a bit of commented out code around - I'd suggest you delete it - if you really need to get it back, your git history still has a copy, but for now it's mostly just adding noise to your code :)

src/App.js
```diff
@@ -1,13 +1,25 @@
...
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        addresses={[
```
illicitonion, line 17: This works, but it'd be good to think about the naming of this prop - the things here aren't all addresses - can you think of a different name that may be a little more accurate?

src/TouristCards.js
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+
+function TouristInfoCards() {
```
illicitonion, line 3: This component does the same thing three times - that works, but can you think of a sub-component you could make to make this less repetitive?

src/Bookings.js
```diff
@@ -1,9 +1,17 @@
...
+//import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
+  const [bookings, setBookings] = useState([]);
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => setBookings(data));
+  }, []);
```
40thieves, line 13: Great 🙂

src/Order.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order(props) {
+  const [orders, setOrders] = useState(0);
+  function orderOne() {
+    setOrders(orders + 1);
```
40thieves, line 7: Well done! 👍

src/Restaurant.js
```diff
@@ -1,14 +1,24 @@
...
+      <ul className="food-orders">
         <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
+          <Order orderType="Pizzas" />
         </li>
+        <li>
+          <Order orderType="Salads" />
+        </li>
+        <li>
+          <Order orderType="Chocolate Cake" />
```
40thieves, line 17: 🍰

src/SearchResults.js
```diff
@@ -1,11 +1,22 @@
...
+  // const [clickRow, setClickRow ] = useState(true) - l16 try?
+
+  // function RowClick(){
+  //   if (clickRow===true){
+  //   {'tr {background-color: red; }'}
+  //     setClickRow(false)
+  //   } else {
+  //     setClickRow(true)
+  //   }
+  // }
```
40thieves, line 14: This is a good attempt, but it looks like you realised that if you click on any row, then all of them become highlighted.

We'll look at a simple solution on Sunday, but the idea is to think carefully about how we could split up our components so that they can have their own state. Can you figure out how to fix it?

PR #101, opened 2020-05-23
Alisina's Hotel Project 7 levels complete

Alisina92:


Alisina92:
[https://cyf-alisina-hotelproj.netlify.app/](url)

src/App.js
```diff
@@ -1,13 +1,23 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        Address={[
```
illicitonion, line 15: I'd maybe call this "addresses" (because it's an array) or even "contact_details" (because they're not all addresses), rather than "Address" :)

src/SearchButton.js
```diff
@@ -0,0 +1,13 @@
+import React from "react"; // imported react library
+
+/*extracted the search button from the search on a new component called search Button */
+const SearchButton = () => {
+  return (
+    // returned the button with JSX tags in the  search button component
+    <div>
```
illicitonion, line 7: Why have you got this `div` here? It's not wrong, but I don't think it's necessary either...

src/TouristInfoCards.js
```diff
@@ -0,0 +1,63 @@
+/*Level 3  */
+
+import React from "react";
+
+const TouristInfoCards = () => {
+  // TouristInfoCards
+  return (
+    /* JSX code that holds a div called section  and an article which holds card info */
+    <div className="section">
+      <article className="card">
```
illicitonion, line 10: This works, but it's a little repetitive, e.g. you're having to say three times that a card has an `h3` element in it, once for each card. Could you try extracting a `Card` component to remove that duplication?

Alisina92: Thank you, Daniel. I am thinking to put the info in JSON and map though will this work?

40thieves:
@Alisina92 I can see that you've pushed new commits to this PR, but you haven't removed the "reviewed" label. Is the week 2 homework complete? If you look in Google Classroom, there's extra instructions on how to remove the "reviewed" label.

PR #103, opened 2020-05-23
Finished lesson1

cockochan:
Vry happy


src/Card.js
```diff
@@ -0,0 +1,14 @@
...
+      <img src={props.city.image} className="card-img-top" alt=" Glasgow" />
+      <div className="card-body">
+        <a href={props.city.href} className="btn btn-primary">
+          Go to {props.city.name}
+        </a>
+      </div>
+    </ul>
+  );
+};
+export default Card;
```
mozartdiniz, line 14: Nice component! Small and clean.

src/Footer.js
```diff
@@ -0,0 +1,13 @@
...
+const Footer = props => {
+  return (
+    <ul className="footerAdress">
+      {props.contactInfo.map((line, index) => (
+        <li key={index}>{line}</li>
+      ))}
+    </ul>
+  );
+};
+export default Footer;
```
mozartdiniz, line 13: Good one.

src/Heading.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+function Heading(props) {
+  const Heading = <header className="App-header">CYF Hotel</header>;
+  return Heading;
+}
+
+export default Heading;
```
mozartdiniz, line 7: Pro tip:

In this case, when the component is just one line, you could do this way.
```js
function Heading(props) {
  return <header className="App-header">CYF Hotel</header>;
}
```

mozartdiniz: Another point is. Are you planning to receive any prop in here? If you don't want to make this component dynamic and reusable, you can just take the props of your function declaration.

You can do:
```js
function Heading() {
```

But, maybe would be better to receive the description as a prop. This way you could reuse this component in other parts of your app and provide different texts.

src/SearchButton.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+function SearchButton() {
+  return (SearchButton = <button className="btn btn-primary">Search</button>);
+}
```
mozartdiniz, line 4: Ok, in the code above, you are providing something that I think is a callback, via props, and naming the prop `SearchButton`, right?

Where inside the SearchButton are you receiving the props? Where inside the SearchButton are you using the callback you are providing to it?

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import fakeBookings from "./fakeBookings.js";
+
+import Tbody from "./Tbody.js";
+// console.log(fakeBookings);
```
mozartdiniz, line 5: Please don't send commented code when submit pull requests.

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
...
+          <th scope="col">days stayed</th>
+        </tr>
+      </thead>
+      <Tbody />
+    </table>
+  );
+};
+
+export default SearchResults;
+/* {data.map((custumerHistory, index)=>
```
mozartdiniz, line 28: Please don't send commented code when submit pull requests.

src/Tbody.js
```diff
@@ -0,0 +1,33 @@
...
+const Tbody = () => {
+  return (
+    <tbody>
+      {fakeBookings.map(booker => {
+        console.log(fakeBookings);
+
+        //         var a = moment([2007, 0, 29]);
+        // var b = moment([2007, 0, 28]);
+        // a.diff(b, 'days') // 1
+        let b = moment(booker.checkInDate);
```
mozartdiniz, line 13: you can use const in here.

src/Tbody.js
```diff
@@ -0,0 +1,33 @@
...
+        let a = moment(booker.checkOutDate);
```
mozartdiniz, line 14: you can use const in here.

src/Tbody.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import fakeBookings from "./fakeBookings.js";
+import moment from "moment";
+const Tbody = () => {
+  return (
+    <tbody>
+      {fakeBookings.map(booker => {
+        console.log(fakeBookings);
```
mozartdiniz, line 8: I think you forgot some debugging code in here.

src/Tbody.js
```diff
@@ -0,0 +1,33 @@
...
+
+        //         var a = moment([2007, 0, 29]);
+        // var b = moment([2007, 0, 28]);
```
mozartdiniz, line 11: No commented code, please. 😄

src/Tbody.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import fakeBookings from "./fakeBookings.js";
```
mozartdiniz, line 2: Ok, I will let this question for you. Should this component import `fakeBookings` directly, o should receive it via props?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,12 @@
...
+function TouristInfoCards(props) {
+  return (
+    <div className="cardsBlock">
+      {props.cities.map((city, index) => (
+        <Card city={city} key={index} />
+      ))}
+    </div>
+  );
+}
+export default TouristInfoCards;
```
mozartdiniz, line 12: Nice!

mozartdiniz: Where are you using this component? I can't find.

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
...
+          <th scope="col">surname</th>
+          <th scope="col">email</th>
+          <th scope="col">room id</th>
+          <th scope="col">check in date</th>
+          <th scope="col">check out date</th>
+          <th scope="col">days stayed</th>
+        </tr>
+      </thead>
+      <Tbody />
+    </table>
```
mozartdiniz, line 23: This is totally ok. But, don't you think would make more sense to have a <BookingsTable /> component rather than a component with just the <Tbody />?

src/Search.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
+import SearchButton from "./SearchButton.js";
 
-const Search = () => {
+const Search = props => {
```
mozartdiniz, line 4: Where is this <Search /> component been used?

src/Tbody.js
cockochan: I expect it to be imported directly, in fact i had both versions, first i was sending it from another component and in the end i thought that just importing it is safer and all was still working  and i did not mind to check the logic from the beginning to end

src/TouristInfoCards.js
cockochan: that should be a container for all of the cards in App.js let me check if i can find it

src/SearchResults.js
cockochan: i see mistake thread, may be fixing that will make my app better.

src/Search.js
cockochan: on line 21 i think

src/SearchResults.js
cockochan: thank you, i did not know will work on that

src/SearchResults.js
cockochan: i found this part very confusing especially splitting the table which i copied with bootstrap styles and break it in two parts was difficult. i was not  expecting some things to have perfect logic as long as it was working,i required help at this point

src/TouristInfoCards.js
cockochan: it's on line 39 in App.js ```<TouristInfoCards cities={cities} />```

src/SearchButton.js
cockochan: As i remember "className="btn btn-primary" is a provided line of code which i should change from html into JSX but keep the styles. I just kept thing i don't understand for safety

src/SearchResults.js
cockochan: probably having just <th> component would be cleaner, but at that moment required additional logic, i thought about exporting object values and keys into separate arrays and then mapping  into <th> and had no simpler way of doing that.

src/Tbody.js
cockochan: at this moment i prefer  original unchanged data, but on later steps if i will be filtering the array before creating cards it's good to have it not imported directly but stored in variable

src/Tbody.js
cockochan: yes

src/SearchResults.js
mozartdiniz: Ok, let’s try to find some time this week and try to do it together. What you think?

cockochan:
https://cyf-cockochan-hotel-react.netlify.app/

cockochan:
new one

src/Bookings.js
```diff
@@ -1,22 +1,26 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
 import SearchResults from "./SearchResults.js";
 // import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import fakeBookings from "./data/fakeBookings.json";
```
40thieves, line 5: Is this fake data being used somewhere? If not I'd recommend removing it so it doesn't confuse you.

src/Bookings.js
```diff
@@ -1,22 +1,26 @@
...
-    console.info("TO DO!", searchVal);
-  };
+  const [bookings, setBookings] = useState(fakeBookings);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
+      .then(res => res.json())
+      .then(data => {
+        setBookings(data);
+      });
+  }, []);
```
40thieves, line 15: Well done 👍

src/Order.js
```diff
@@ -0,0 +1,15 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+const Order = props => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = props => {
+    setOrders(orders + 1);
```
40thieves, line 6: Nice work! 🎆

src/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
+import Order from "./Order";
+const Restaurant = props => {
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order {...props} orderType="pizzas" />
```
40thieves, line 9: I don't think it causes a problem, but what props are we receiving in this component?

I don't think the *spread* (`{...props}`) is necessary?

src/Tbody.js
```diff
@@ -1,15 +1,10 @@
 import React from "react";
-import fakeBookings from "./fakeBookings.js";
 import moment from "moment";
-const Tbody = () => {
+const Tbody = props => {
+  console.log(props);
```
40thieves, line 4: Did you attempt exercise 16? I think you'll need to update this component a bit to be able to do that exercise.

PR #104, opened 2020-05-24
React-week1-( HotelApp )

kiowd:


src/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const Footer = props => {
+  const addresses = props.address;
+  const mainAddress = addresses.map((info, index) => (
+    <li key={index}>{info}</li>
+  ));
+  return <ul className="footer">{mainAddress}</ul>;
```
40thieves, line 8: A very minor point, but you could make this a bit shorter:

```js
const mainAddress = props.addresses.map((info, index) => (
  <li key={index}>{info}</li>
));
return <ul className="footer">{mainAddress}</ul>;
```

And you could make that even shorter:

```js
return (
  <ul className="footer">
    {props.addresses.map((info, index) => (
      <li key={index}>{info}</li>
    ))}
  </ul>
);
```

src/App.js
```diff
@@ -1,13 +1,25 @@
...
+];
+
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer address={addresses} />
```
40thieves, line 22: 👍

src/SearchButton.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+const SearchButton = () => {
+  return <button className="btn btn-primary">Search</button>;
```
40thieves, line 4: Nice work 🙂

src/SearchResults.js
```diff
@@ -0,0 +1,40 @@
...
+    <tr key={index}>
+      <th scope="row">{user.id}</th>
+      <td>{user.title}</td>
+      <td>{user.firstName}</td>
+      <td>{user.surname}</td>
+      <td>{user.email}</td>
+      <td>{user.roomId}</td>
+      <td>{user.checkInDate}</td>
+      <td>{user.checkOutDate}</td>
+      <td>{moment(user.checkOutDate).diff(user.checkInDate, "days")} days</td>
```
40thieves, line 15: Great work 💪

src/SearchResults.js
```diff
@@ -0,0 +1,40 @@
...
+            <th scope="col">first name</th>
+            <th scope="col">surname</th>
+            <th scope="col">email</th>
+            <th scope="col">room id </th>
+            <th scope="col">check in date</th>
+            <th scope="col">check out data</th>
+            <th scope="col">number of nights</th>
+          </tr>
+        </thead>
+        <tbody className="result">{rows}</tbody>
```
40thieves, line 34: Similar to above, this could be "inlined" by inserting the code assigned to `rows` here.

But this is fine too! And sometimes I prefer this approach 🙂

src/TouristInfoCards.js
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+
+const TouristInfoCards = () => {
```
40thieves, line 3: This component is responsible for rendering all of the Tourist Info cards.

Could instead make a component that is responsible for rendering a single card? What props would you pass to this component?

src/SearchResults.js
kiowd: Thank you very much for that and I did come across I can use "inlined" when I started working on High-Sores-Table project. :)

src/TouristInfoCards.js
kiowd: I should make an array of objects to store those data and pass it as prop to TouristInfoCards and loop through it to render by using only a single card.

src/Bookings.js
```diff
@@ -1,9 +1,17 @@
...
 const Bookings = () => {
+  const [bookings, setBookings] = useState([]);
+  const URL = "https://cyf-react.glitch.me";
+
+  useEffect(() => {
+    fetch(URL)
+      .then(response => response.json())
+      .then(data => setBookings(data));
+  }, []);
+
```
40thieves, line 14: Nice 👍

src/Order.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = ({ orderType }) => {
+  const [orders, setOrders] = useState(0);
+
+  const orderOne = () => {
+    setOrders(orders + 1);
```
40thieves, line 8: Great work!

src/Restaurant.js
```diff
@@ -1,14 +1,15 @@
...
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+
+      <ul className="list-group">
+        <Order orderType="Pizzas" />
+        <Order orderType="Salads" />
+        <Order orderType="Chocolate cake" />
```
40thieves, line 12: 🍰

src/SearchResults.js
```diff
@@ -1,9 +1,21 @@
...
+import React, { useState } from "react";
 import moment from "moment";
 
+import "./App.css";
+
 const SearchResult = props => {
+  const [active, setActive] = useState(null);
+
+  const handleClick = index => {
+    active === index ? setActive(null) : setActive(index);
```
40thieves, line 10: This is a good attempt, but it looks like if you click on any row, then all of them become highlighted. The exercise asks you to make each row *independent*, meaning that multiple rows can be highlighted at once.

The idea is to think carefully about how we could split up our components so that they can have their own state. Can you work out how to do that?

PR #106, opened 2020-05-24
Cyf react hotel min ko

minwwgo:


src/App.js
```diff
@@ -1,13 +1,25 @@
...
+const name = " CYF hotel";
+
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Header name={name} />
+      <TouristInfoCards data={Locationinfo} />
       <Bookings />
+      <Footer address={address} />
```
illicitonion, line 22: "address" is a slightly strange name to use for this prop, both because it's not just one address, and because not everything in the array is an address. Can you think of a better name?

src/Search.js
```diff
@@ -17,7 +17,8 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
+              <SearchButton />
+              {/* <button className="btn btn-primary">Search</button> */}
```
illicitonion, line 21: Why keep the commented out code?

src/SearchButton.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+const SearchButton = () => {
+  return (
+    <div>
```
illicitonion, line 5: Why do you have a div around this button? It works, but I don't think it's necessary... Can you talk through why you added it?

minwwgo: ya I will take it out

src/App.js
minwwgo: I changed to contactInfos

src/SearchResults.js
```diff
@@ -1,9 +1,22 @@
...
 import moment from "moment";
 
 const SearchResults = props => {
+  const [rowSelect, setRowSelect] = useState("");
+
+  const changeBackground = data => {
+    if (data === rowSelect) {
+      setRowSelect();
+    } else {
+      setRowSelect(data);
```
40thieves, line 11: This is a really good attempt, but it looks like if you click on any row, then all of them become highlighted. The exercise asks you to make each row *independent*, meaning that multiple rows can be highlighted at once.

We'll look at a simple solution on Sunday, but the idea is to think carefully about how we could split up our components so that they can have their own state.

40thieves:
Hey @minwwgo it also looks like your Hotel app is not deployed at the correct Netlify URL: https://cyf-minwwgo-hotel-react.netlify.app/. Could you make sure that's fixed?

PR #107, opened 2020-05-24
Melanie's React hotel project

melanieet:


src/App.js
```diff
@@ -2,12 +2,31 @@ import React from "react";
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards
+        city="Glasgow"
+        site="https://peoplemakeglasgow.com/"
+        image="https://frontline-collections.com/wp-content/uploads/2018/02/Glasgow-Debt-Collection.jpg"
+      />
+      <TouristInfoCards
```
mozartdiniz, line 18: Good job. I like the way you created this component and is reusing it. What if we created a data structure which we could map and return this component? Wouldn't be cool? We could avoid copy and paste this component over and over again.

Something like:
```js
const touristInformations = [{
   city: 'Glasgow',
   site: 'https://peoplemakeglasgow.com/',
   imageURL: '"https://frontline-collections.com/wp-content/uploads/2018/02/Glasgow-Debt-Collection.jpg'
}, {
   city: 'Manchester',
   site: 'https://www.visitmanchester.com/',
   imageURL: 'https://www.prolificnorth.co.uk/sites/default/files/styles/lightbox_large/public/images/news/manchester-cityscape0.jpg?itok=VjQDdTYJ'
}]
```

What do you think?

src/Footer.js
```diff
@@ -0,0 +1,21 @@
...
+
+const Footer = () => {
+  return (
+    <footer className="footer">
+      <ul>
+        {info.map((element, index) => (
+          <li key={index}>{element}</li>
+        ))}
+      </ul>
+    </footer>
```
mozartdiniz, line 17: This is nice! What if we received the info via props? Wouldn't make this component more dynamic?

src/SearchResults.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import FakeBookings from "./data/fakeBookings.json";
```
mozartdiniz, line 2: You should only capitalize the first letter of the variable if it is:
- A React component
- A class

In this case, would be better if you called it fakeBookings.

src/SearchResults.js
```diff
@@ -0,0 +1,49 @@
...
+import moment from "moment";
+
+const SearchResults = () => {
+  return (
+    <div>
```
mozartdiniz, line 7: This is really cool. I like the way you created the table and use map to the details.

One thing, let's imagines we are going to get those FakeBookings from somewhere outside of this component. From a component above, or even load from the internet. In both cases, it would come from outside. What do you think if we receive those bookings as props? I think would make it more dynamic.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,16 @@
...
+      <div className="card-body">
+        <a href={props.site} className="btn btn-primary">
+          Visit {props.city}
+        </a>
+      </div>
+    </div>
+  );
+};
+
+export default TouristInfoCards;
```
mozartdiniz, line 16: Nice small and clean component.

PR #109, opened 2020-05-25
cynthia-hotel-react-lesson1

cyeburu:
Lesson1 - Done

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,19 @@
 import React from "react";
+import Heading from "./Heading"
+import TouristInfoCards from "./TouristInfoCards"
+import Footer from "./Footer"
 
 import Bookings from "./Bookings";
 import "./App.css";
 
 const App = () => {
+  const footerContentArray = ["123 Fake Street, London, E1 4UD", "hello@fakehotel.com", "0123 456789"]
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer footercontent={footerContentArray} />
     </div>
   );
 };
--- a/src/Search.js
+++ b/src/Search.js
@@ -1,4 +1,6 @@
 import React from "react";
+import SearchBtn from "./SearchBtn"
+
 
 const Search = () => {
   return (
@@ -10,15 +12,11 @@ const Search = () => {
         <div className="col">
           <form className="form-group search-box">
             <label htmlFor="customerName">Customer name</label>
-            <div className="search-row">
-              <input
-                type="text"
-                id="customerName"
-                className="form-control"
-                placeholder="Customer name"
-              />
-              <button className="btn btn-primary">Search</button>
-            </div>
+            <label className="searchIdLabel" htmlFor="customerName">Customer id</label>
+          </form>
+
+          <form className="form-group search-box">
+            <SearchBtn />
           </form>
         </div>
       </div>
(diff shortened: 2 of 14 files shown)


src/App.js
```diff
@@ -1,13 +1,19 @@
...
 
 const App = () => {
+  const footerContentArray = ["123 Fake Street, London, E1 4UD", "hello@fakehotel.com", "0123 456789"]
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer footercontent={footerContentArray} />
```
illicitonion, line 16: Because this `prop` is specific to the `Footer`, it's kind of unnecessary to call it `footercontent` - we could just call it `content` (we're giving it to the `Footer`, so we don't need to say `Footer` again).

But in _this_ component, `footerContentArray` is a good variable name, because _this_ component deals with many child components.

src/Search.js
```diff
@@ -10,15 +12,11 @@ const Search = () => {
...
-            <div className="search-row">
-              <input
-                type="text"
-                id="customerName"
-                className="form-control"
-                placeholder="Customer name"
-              />
-              <button className="btn btn-primary">Search</button>
-            </div>
+            <label className="searchIdLabel" htmlFor="customerName">Customer id</label>
```
illicitonion, line 15: These `label` elements relate to the `input` elements in the `SearchBtn` component. This is a little strange - ideally components don't need to know about the internals of each other, but right now they do - these `label` elements know what the `id` attributes of things in the `SearchBtn` component are.

I would recommend moving the `input` elements back into this component, and having the `SearchBtn` just be the button. This has the advantage of making `SearchBtn` also be accurately named (Right now it is more than just a button!)

Additionally, if you wanted to remove some duplication, we could extract a new component, `SearchBox` which has as children a `label`, and `input`, and a `SearchBtn`.

Why don't you give that a go, it'll be good practice at extracting components :)

src/SearchResults.js
```diff
@@ -0,0 +1,45 @@
...
+                    <th scope="col">checkInDate</th>
+                    <th scope="col">checkOutDate</th>
+                    <th scope="col">totalNight</th>
+                </tr>
+            </thead>
+            <tbody>
+                {props.results.map((bookingData, index) => {
+                    return (
+                        <tr key={index} >
+                            <th key={"id" + index}>{bookingData.id}</th>
```
illicitonion, line 26: As long as the `tr` element has a `key`, I don't think its children need keys, though it doesn't hurt to have them here :)

src/TotalNightsCol.js
```diff
@@ -0,0 +1,13 @@
+import React from "react"
+import SearchResults from "./SearchResults"
+
+
+const TotalNightsCol = () => {
```
illicitonion, line 5: You don't use this component - maybe delete it?

src/Search.js
```diff
@@ -14,10 +14,23 @@ const Search = () => {
...
             <label className="searchIdLabel" htmlFor="customerName">Customer id</label>
           </form>
+          <div className="search-row">
+            <input
+              type="text"
+              id="customerName"
+              className="form-control"
+              placeholder="Customer name"
+            />
+            <button className="btn btn-primary">Search</button>
```
illicitonion, line 24: This is exactly the same as the `SearchBtn` component - let's re-use that, so that if we change it in the future (e.g. give it a new class), we only need to change it in one place :)

PR #110, opened 2020-05-25
Roxana: hotel-react project lesson 1

279Roxana:


src/Search.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
+import SearchButton from "./SearchButton"
 
-const Search = () => {
+const Search = (props) => {
```
illicitonion, line 4: It's a little strange, but not wrong, to take props and then not use them - in your other components where you don't use props, you just used `()` as the argument list :)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+const TouristInfoCards = (props) => {
+  return( 
+    <div className="tourist-cards">
+
+        {props.cardAttributes.map((attribute, index) => (
+            <div className="card" key ={index}>
+    <img src={attribute.img} className="card-img-top" alt= "{attribute.city} city" />
```
illicitonion, line 9: We'd normally indent things so that children (like this `img` tag) are further right than their parents (the containing `div` tag) - doesn't change the functionality, but makes it a little easier to read the code :)

src/Search.js
```diff
@@ -10,15 +11,7 @@ const Search = () => {
         <div className="col">
           <form className="form-group search-box">
             <label htmlFor="customerName">Customer name</label>
```
illicitonion, line 13: The `label` here is kind of attached to the `input` element that you moved into the component. At this point, it doesn't really matter whether the `input` is in the `SearchButton` component or the `Search` component (it may later :)), but the `label` and the `input` should definitely be in the _same_ component, because one should not be present without the other.

PR #111, opened 2020-05-25
Hotel react

Hamidreza-Fakharzadeh:


Hamidreza-Fakharzadeh:
updated

Hamidreza-Fakharzadeh:
updated

src/Footer.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+const arrAddress = [
+  {
+    address: "123 Fake Street, London, E1 4UD",
+    email: "hello@fakehotel.com",
+    tel: "0123 456789"
+  }
+];
+
```
40thieves, line 10: If you read the instructions for exercise 4 carefully, you'll see that it asks you to pass this data *as a prop* to the `Footer` component. Could you change this component so it does that?

src/Heading.js
```diff
@@ -0,0 +1,21 @@
...
+function Heading() {
+  return (
+    <header className="App-header">
+      {" "}
+      CYF Hotel <Logo />
+    </header>
+  );
+}
+
+function Logo() {
```
40thieves, line 12: Normally we like to have every separate component in it's own file. Where do you think we should move this code to?

src/SearchButton.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function SearchButton() {
+  return <button className="btn btn-primary"> Search </button>;
```
40thieves, line 4: Nice work!

src/SearchResults.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import moment from "moment";
+
+function SearchResults(props) {
+  function calculateDuration(checkOutDate, checkInDate) {
+    return moment(checkOutDate.toString()).diff(
+      moment(checkInDate.toString()),
+      "days"
+    );
+  }
```
40thieves, line 10: Great :)

src/SearchResults.js
```diff
@@ -0,0 +1,49 @@
...
+          <th>Room</th>
+          <th>Check In</th>
+          <th>Check out</th>
+          <th>Duration</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map((result, index) => {
+          return (
+            <tr key={index}>
```
40thieves, line 29: Well done 🍰

src/TouristInfoCards.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+function TouristInfoCards() {
```
40thieves, line 3: This component renders all 3 tourist info cards. But if I wanted to show just 1 of the cards in another place in the app, I'd have to copy this code to where I wanted it.

src/backupSearch.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
40thieves, line 1: It's good to delete code if it's not being used. Otherwise you could confuse yourself about what code is actually being used.

Hamidreza-Fakharzadeh:
Hi Alasdair, I really appreciate your review on my code and your
suggestions. I will correct it
as you suggested.
Thank you

Hamidreza-Fakharzadeh:
finished lesson2

Hamidreza-Fakharzadeh:
updated

Hamidreza-Fakharzadeh:
Thanks to all for reviewing my code.

Hamidreza-Fakharzadeh:
Thanks to all for reviewing my code.

Hamidreza-Fakharzadeh:
finished task

Hamidreza-Fakharzadeh:
Thank you for your review.

Hamidreza-Fakharzadeh:
I set Api for get request to grab data by id

Hamidreza-Fakharzadeh:
I set Api for get request to grab data by id

Hamidreza-Fakharzadeh:
resolved conflict

Hamidreza-Fakharzadeh:
updated

PR #112, opened 2020-05-25
lesson 1-completed

mahmutsel:


src/App.css
```diff
@@ -55,3 +60,22 @@ tr {
...
+  height: auto;
+}
+/* .cards {
+  display: flex;
+  align-items: center;
+  margin: 0 auto;
+  padding: 10px 10px;
+} */
+.cards {
+  display: flex;
```
40thieves, line 77: Nice use of flexbox!

src/App.js
```diff
@@ -1,13 +1,26 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <div className="footer footer-list">
+        <Footer contactDetails={details} />
```
40thieves, line 22: Excellent work!

src/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
```
40thieves, line 4: A very minor problem, but normally we only use capital letters in the first letter for React components. Everything else we name like a normal variable, so the first letter is not capitalised.

Could you fix this so it follows that *convention*?

src/Search.js
```diff
@@ -17,9 +18,9 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
             </div>
           </form>
+          <SearchButton />
```
40thieves, line 23: I don't think this will cause a big problem, but I just wanted to point out that the new `SearchButton` component has moved slightly from where it was before.

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+        ))}
+      </tbody>
+    </table>
+  );
+};
+
+function daysDifference(start, end) {
+  const duration = moment.duration(moment(start).diff(moment(end))).asDays();
+  return Math.round(duration);
+}
```
40thieves, line 42: Excellent!

I think another solution to this problem would be to do something like this:

```js
let startDate = moment(start)
let endDate = moment(end)
let diffInDays = startDate.diff(endDate, "days")
```

Or as you noticed, we could *inline* some of those variables:

```js
moment(start).diff(moment(end), "days")
```

src/TouristInfoCards.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+function TouristInfoCards() {
```
40thieves, line 3: This component renders all 3 tourist info cards. But if I wanted to show just 1 of the cards in another place in the app, I'd have to copy this code to where I wanted it.

mahmutsel: It makes sense :+1:

src/Bookings.js
```diff
@@ -1,19 +1,24 @@
...
+  const [bookings, setBookings] = useState([]);
+
   const search = searchVal => {
     console.info("TO DO!", searchVal);
   };
-
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
+      .then(res => res.json())
+      .then(data => setBookings(data));
```
40thieves, line 14: Nice work! 👏

src/Order.js
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order(props) {
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    setOrders(orders + 1);
```
40thieves, line 8: Really clean component! 👏

PR #113, opened 2020-05-26
abcelen-cyf-hotel

abcelen:
Lesson 1 complete

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,26 @@
 import React from "react";
-
+import Heading from "./Heading";
+import TouristInfoCards from "./TouristInfoCards";
 import Bookings from "./Bookings";
+import Footer from "./Footer";
+// import searchButton from "./searchButton";
 import "./App.css";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      {/* <header className="App-header">CYF Hotel</header> */}
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        footerProps={[
+          "123 Fake Street, London, E1 4UD",
+          "hello@fakehotel.com",
+          "0123 456789"
+        ]}
+      />
+      />
     </div>
   );
 };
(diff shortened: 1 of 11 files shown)


src/App.js
```diff
@@ -1,13 +1,26 @@
...
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      {/* <header className="App-header">CYF Hotel</header> */}
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        footerProps={[
```
40thieves, line 17: Do you think this is a good name for the prop? We already know that it is a prop and that it's being passed to the Footer component, so it's not really helping us understand what this prop is for.

src/Search.js
```diff
@@ -17,7 +18,7 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
+              <SearchButton />
```
40thieves, line 21: 👍 🍰

src/SearchResults.js
```diff
@@ -0,0 +1,47 @@
...
+            <tr>
+              <td>{data.id}</td>
+              <td>{data.title}</td>
+              <td>{data.firstName}</td>
+              <td>{data.surname}</td>
+              <td>{data.email}</td>
+              <td>{data.roomId}</td>
+              <td>{data.checkInDate}</td>
+              <td>{data.checkOutDate}</td>
+              <td>{checkIn.diff(checkOut, "days")}</td>
```
40thieves, line 34: Great work!

src/TouristInfoCards.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+const TouristInfoCards = () => {
```
40thieves, line 2: This component renders all 3 tourist info cards. But if I wanted to show just 1 of the cards in another place in the app, I'd have to copy this code to where I wanted it.

PR #114, opened 2020-05-28
Lesson 1 complited

AlexeyBrambalov:


src/App.js
```diff
@@ -1,13 +1,23 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        list={[
```
40thieves, line 15: Is this the best name for this prop? Does it describe what is actually being passed?

src/components/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <ul className="bg-dark text-white-50 text-center list">
+      {props.list.map((item, index) => (
+        <li key={index}>{item}</li>
```
40thieves, line 7: Great 👍

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+function TouristInfoCards() {
```
40thieves, line 3: This component renders all 3 tourist info cards. But if I wanted to show just 1 of the cards in another place in the app, I'd have to copy this code to where I wanted it.

src/App.js
AlexeyBrambalov: yes, I updated it

src/components/TouristInfoCards.js
AlexeyBrambalov: yes, I updated it as a new component TouristInfoCard

PR #115, opened 2020-05-28
Babak-Class6 - Lesson 3 done (forgot to pull Req) !

babakjahan:
thank you


src/Restaurant.js
```diff
@@ -1,13 +1,23 @@
-import React from "react";
+import React, { useState } from "react";
 
 const Restaurant = () => {
-  const pizzas = 0;
+  const [orders, setOrders] = useState(0);
```
40thieves, line 4: Great that you've already started to look at this 🙂 We'll look at this in more detail on Sunday

src/Search.js
```diff
@@ -11,12 +12,7 @@ const Search = () => {
           <form className="form-group search-box">
             <label htmlFor="customerName">Customer name</label>
             <div className="search-row">
-              <input
-                type="text"
-                id="customerName"
-                className="form-control"
-                placeholder="Customer name"
-              />
+              <SearchButton />
```
40thieves, line 15: This isn't quite what what exercise 1 asks you to do: it says you should move the `<button>` element to a separate component, not the `<input>`. Could you fix it?

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import "./css/Footer.css";
+import FakeBookingData from "./data/fakeBookings.json";
```
40thieves, line 3: If you read the instructions in exercise 6 carefully it says that you should pass the fake bookings data from the `Bookings` component as a prop to the `SearchResults` component. Can you fix that please?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,59 @@
+import React from "react";
+import "./css/TouristInfoCards.css";
+import londonPic from "./images/ london-web.jpg";
+import glasgowPic from "./images/glasgow-web.jpg";
+import manchesterPic from "./images/manchester_web.jpg";
+
+const TouristInfoCard = props => {
```
40thieves, line 7: It's really great that decided to split out the `TouristInfoCard` component! It makes your code much more flexible :)

But remember that we always put components in separate files so that we don't get confused. Could you move this to a separate file?

40thieves (commented):
Nice work @babakjahangiri! 🍰 🎆 👍

src/Search.js
babakjahangiri: done

src/SearchResults.js
babakjahangiri: exercise 6 issues have resolved

src/BookingRow.js
```diff
@@ -1,13 +1,27 @@
...
+import React, { useState } from "react";
 import moment from "moment";
 
 function calcDeff(chkIn, chkOut) {
   return moment(chkOut, "YYYY-MM-DD").diff(moment(chkIn, "YYYY-MM-DD"), "days");
 }
 
 const BookingRow = props => {
+  const [bookingRowClass, setBookingRowClass] = useState("");
+  const [selected, setSelected] = useState(false);
```
40thieves, line 10: Really excellent work! Well done 🙂 Can you explain how this solves the problem in exercise 16?

src/BookingRow.js
```diff
@@ -1,13 +1,27 @@
...
+
+  function selectRow() {
+    // console.log("someone calls me");
+    if (selected === false) {
+      setSelected(true);
+      setBookingRowClass("search-selected");
```
40thieves, line 16: This is correct, but some good React advice is to use as little state as possible. So here, we can calculate what the `bookingRowClass` should be based on the `selected` state. This means that if we ever want to change the class to something different, we only need to change it in 1 place. Does that make sense?

src/Bookings.js
```diff
@@ -1,18 +1,26 @@
...
+      .then(res => res.json())
+      .then(data => SetBookings(data));
+  }, []);
+
   return (
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults BookingsData={bookings} />
```
40thieves, line 23: The React *convention* is to make prop names start with a lower-case letter. Only components should start with an upper-case letter. Could you fix that here?

src/Bookings.js
```diff
@@ -1,18 +1,26 @@
...
     console.info("TO DO!", searchVal);
   };
 
+  const [bookings, SetBookings] = useState([]);
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => SetBookings(data));
+  }, []);
```
40thieves, line 17: Great work! 🏅

src/Order.js
```diff
@@ -0,0 +1,18 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = props => {
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    setOrders(orders + 1);
```
40thieves, line 8: This is a really nice and clean component 👏

src/Restaurant.js
```diff
@@ -1,24 +1,14 @@
...
-          <button
-            onClick={() => setOrders(orders + 1)}
-            className="btn btn-primary"
-          >
-            Add
-          </button>
-        </li>
+        <Order orderType="Pizzas" />
+        <Order orderType="Salads" />
+        <Order orderType="Chocolate cake" />
```
40thieves, line 11: 🍰

src/SearchResults.js
```diff
@@ -1,42 +1,47 @@
...
+              <th scope="col">Check-out</th>
+              <th scope="col">Duration</th>
+            </tr>
+          </thead>
+          <tbody>
+            {props.BookingsData.map((booking, index) => {
+              return (
+                <BookingRow
+                  key={booking.id}
+                  id={booking.id}
```
40thieves, line 30: I think this is probably fine, but there is a way we could pass many fewer props. How do you think we could do that?

40thieves (commented):
Really nice work @babakjahangiri! ✨ 👍 🎉

PR #120, opened 2020-05-31
Yohannes Hotel react Lesson 1

byohannes:


src/App.js
```diff
@@ -1,13 +1,23 @@
...
+];
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Restaurant />
+      <Footer arrayInfo={info} />
```
40thieves, line 20: Do you think `arrayInfo` is the most descriptive name? Does it help us understand what is in that prop?

src/Bookings.js
```diff
@@ -1,18 +1,27 @@
...
     console.info("TO DO!", searchVal);
   };
 
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => setBookings(data));
+  }, []);
+  if (!bookings) {
+    return null;
```
40thieves, line 17: This is some really nice loading/error handling! 👏

src/Bookings.js
```diff
@@ -1,18 +1,27 @@
...
+  const [bookings, setBookings] = useState([]);
   const search = searchVal => {
     console.info("TO DO!", searchVal);
   };
 
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => setBookings(data));
+  }, []);
```
40thieves, line 15: Well done 👍

src/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order orderType="Pizzas" />
+        <Order orderType="Salads" />
+        <Order orderType="Chocolate cake" />
```
40thieves, line 11: 🍰

src/components/CityCards.js
```diff
@@ -0,0 +1,35 @@
+import londonImage from '../images/london.jpg';
+import glasgowImage from '../images/glasgow.jpg';
+import manchesterImage from '../images/manchester.jpg';
+
+const cities = [
```
40thieves, line 5: This is a nice clean way of separating the data, but the exercise asks you to pass this data as a prop to `TouristInfoCards`. Could you update this?

src/components/Order.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = props => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
+    setOrders(orders + 1);
```
40thieves, line 7: Nice work! 🙂

src/components/SearchResults.js
```diff
@@ -0,0 +1,29 @@
...
+          <th scope="col">Email</th>
+          <th scope="col">Room #</th>
+          <th scope="col">Check-in</th>
+          <th scope="col">Check-out</th>
+          <th scope="col">Duration</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map((customer, index) => {
+          return <TableRow key={index} list={customer} />;
```
40thieves, line 22: Do you think `list` is the best name for this prop? What does it represent? Could we use that as a name instead?

src/components/TableRow.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from "react";
+import moment from "moment";
+const TableRow = ({ list }) => {
+  const checkIn = list.checkInDate;
+  const checkOut = list.checkOutDate;
+  const duration = moment(checkOut).diff(checkIn);
+  const durationInHours = duration / 1000 / 60 / 60;
+  const durationInDays = Math.round(durationInHours / 24);
```
40thieves, line 8: If you look at the Moment documentation, you'll see that the `diff` method actually has a way of calculating this for you: https://momentjs.com/docs/#/displaying/difference/

src/components/TableRow.js
```diff
@@ -0,0 +1,31 @@
...
+
+  const [rowActive, setRowActive] = useState(false);
+
+  const changeRowColor = () => {
+    setRowActive(!rowActive);
+  };
+
+  return (
+    <tr onClick={changeRowColor} className={rowActive ? "row-color" : ""}>
```
40thieves, line 17: This is excellent, well done! 🎉

src/TouristInfoCards .js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
40thieves, line 1: If you look at the name of this file you'll see that you have a space at the end (before the `.js`), which means this file isn't actually being imported and looks like a duplicate. Could you remove it?

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import cities from "./CityCards";
+
+const TouristInfoCards = () => {
+  return (
+    <div className="container">
+      <div className="row">
+        {cities.map((city, index) => (
```
40thieves, line 8: I'm not sure if you talked about this in Group A, but we had a discussion about how it may be useful to further split up this component to show an individual card that uses props to render data.

This way, we can re-use that component all over the application, instead of always having to render multiple cards.

src/App.js
byohannes: I have updated it with a descriptive name such as hotelContactInfo , thanks Ali.

src/Bookings.js
byohannes: Thanks Ali

src/components/CityCards.js
byohannes: I have updated it, thanks.

src/components/SearchResults.js
byohannes: I have replaced it with a good descriptive name, thanks again

src/components/TableRow.js
byohannes: correct,
 const duration = moment(checkOut).diff(checkIn, 'days');

src/TouristInfoCards .js
byohannes: yes, I have sorted thanks.

src/components/TouristInfoCards.js
byohannes: yes, it is a good point and I am considering it, thanks Again .

byohannes: yes, it is a good point and I am considering it, thanks Again .

byohannes:
really appreciating on your detailed review Ali, thanks Again.

PR #124, opened 2020-06-03
Afshin-Hotel-React-Level-1

AKaramifar:


src/App.css
```diff
@@ -1,23 +1,61 @@
+@import url("https://fonts.googleapis.com/css?family=Bebas+Neue|Roboto+Condensed:wght@700|Scada|Squada+One|Nunito+Sans:300,600,800|Concert+One&display=swap");
```
40thieves, line 1: This is a nice touch but you might want to be a bit careful when loading lots of fonts. Font files can be quite large so they can slow your website down a lot.

src/App.js
```diff
@@ -1,13 +1,22 @@
...
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer Info={FooterInfo} />
```
40thieves, line 19: This looks good, but the React *convention* is to start prop names with a lower-case letter. Only component names should start with an upper-case letter.

src/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <div id="Footer" className="Footer_CSS">
+      <p className="Footer_P_CSS">Email Address: {props.Info[1]}</p>
+      <p className="Footer_P_CSS">Tell: {props.Info[2]}</p>
+      <p className="Footer_P_CSS">Address: {props.Info[0]}</p>
```
40thieves, line 8: This works ok, but exercise 4 asks you to use the `map` method to render a list. Could you update this component to use that instead?

src/App.css
```diff
@@ -53,5 +105,112 @@ tr {
...
+  border: none;
+  height: fit-content;
+  padding-bottom: 20px;
+  text-align: center;
+  transition: 200ms;
+}
+.card:hover {
+  transform: scale(1.01);
+}
+.Image_CSS {
```
40thieves, line 119: I'm not sure the `_CSS` bit is necessary? When I'm writing code, I usually only write classes to be used in CSS, so writing it for every class is not needed. What do you think?

src/SearchButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const SearchButton = () => {
+  return (
+    <div id="SearchButton">
+      <button className="btn btn-primary">Search</button>
```
40thieves, line 6: This is good 👍

src/SearchResults.js
```diff
@@ -0,0 +1,45 @@
...
+                  {el.title}
+                </th>
+                <td className="Table_TD_CSS">{el.firstName}</td>
+                <td className="Table_TD_CSS">{el.surname}</td>
+                <td className="Table_TD_CSS">{el.email}</td>
+                <td className="Table_TD_CSS">{el.roomId}</td>
+                <td className="Table_TD_CSS">{el.checkInDate}</td>
+                <td className="Table_TD_CSS">{el.checkOutDate}</td>
+                <td className="Table_TD_CSS">
+                  {moment(el.checkOutDate).diff(moment(el.checkInDate), "days")}
```
40thieves, line 34: Well done! 🙂

src/TouristInfoCards.js
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+
+const TouristInfoCards = () => {
```
40thieves, line 3: I'm not sure if you talked about this in Group A, but in Group B we had a discussion about how it may be useful to further split up this component to show an individual card that uses props to control what the card shows.

This way, we can re-use that component all over the application, instead of always having to render multiple cards.

40thieves:
Remember that when you've pushed the Week 2 homework, you'll need to remove the "reviewed" label.

src/App.css
AKaramifar: You are right @40thieves , I have to remove extra fonts, Thank you so much for review

src/Footer.js
AKaramifar: why I didn't use map because I added label before each item , but I'll change it to map. thank you

src/App.css
AKaramifar: some time our project is big and we have a lots of code I separated id and class with "_CSS" and "_JSX" only for that time that I'm looking to find something this would be helpful for me to find fast and better organizing my code, and this was one of the mentors advice

PR #99, opened 2020-05-23
CYF-HOTEL-PROJECT-TAYFUN

tayfuntayfun:

PR #126, opened 2020-06-05
c-f-hotel-react-tayfun  (PR #99 has been broken )

tayfuntayfun:


src/App.js
```diff
@@ -1,13 +1,28 @@
...
+import Footer from "./Footer";
 
 const App = () => {
+  const countryInfo = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com ",
+    "0123 456789 "
+  ];
+
+  const touristicInformation = TouristicData;
```
40thieves, line 17: We already have a variable `TouristicData`, so do you think another one is necessary?

src/Bookings.js
```diff
@@ -1,18 +1,25 @@
...
     console.info("TO DO!", searchVal);
   };
 
+  const [bookings, setBookings] = useState([]);
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(results => results.json())
+      .then(data => setBookings(data));
+  }, []);
```
40thieves, line 16: Nice work 👍

src/Order.js
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton ";
+
+const Order = props => {
```
40thieves, line 4: This is a really nice and clean component, well done! 👏

src/Restaurant.js
```diff
@@ -1,13 +1,20 @@
...
+      <ul className="order-list">
         <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
+          <Order orderType="Pizzas" />{" "}
+        </li>
+        <li>
+          <Order orderType="Salads" />{" "}
+        </li>
+        <li>
+          <Order orderType="Chocolate cake" />
```
40thieves, line 17: 🍰

src/TableRow.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const Row = ({ booking }) => {
```
40thieves, line 4: Really great job! Can you explain why extracting the table row to a component helps solve the exercise?

src/App.js
tayfuntayfun: DEfinitely not, why I did this :(.

src/Bookings.js
tayfuntayfun: Thanks

src/TableRow.js
tayfuntayfun: Otherwise each row would have only one state and selecting by each would not be possible. Now , each row will be its own state to be updated / to be used for any reason. We can then separate the rows easily .

PR #132, opened 2020-07-12
Orhan-cyf-hotel-react

OBakir90:


src/App.js
```diff
@@ -1,13 +1,24 @@
...
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        info={[
+          "123 Fake Street, London, E1 4UD",
+          "hello@fakehotel.com",
+          "0123 456789"
+        ]}
+      />
     </div>
```
Ashaghel, line 22: Good job here!

src/components/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
-import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import Search from "./Search";
+import SearchResults from "./SearchResults";
+import * as FakeBookings from "../data/fakeBookings.json";
```
Ashaghel, line 4: why you think using * as here is good?, (this is common in TS but not in JS 
Read more 
https://stackoverflow.com/questions/54585763/what-is-the-difference-between-import-as-react-from-react-vs-import-react-fr

src/components/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <div id="footer" className="footer">
+      <ul>
+        {props.info.map((p, i) => {
+          return <li key={i}> {p} </li>;
```
Ashaghel, line 8: good job,

src/components/SearchResults.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import moment from "moment";
+
+const SearchResults = ({ results }) => {
+  return (
+    <table class="table">
+      <thead>
+        <tr>
+          {Object.keys(results.default[0]).map(p => {
+            return <th scope="col">{p}</th>;
```
Ashaghel, line 10: interesting logic !

PR #133, opened 2020-07-13
First lesson complete

Hedyeh65:


src/components/SearchResults.js
```diff
@@ -0,0 +1,17 @@
...
+import TableInfo from "./TableInfo.js";
+import TableData from "./TableData.js";
+import FakeBookings from "./data/fakeBookings.json";
+
+function SearchResults() {
+  return (
+    <div>
+      <table className="table table-striped">
+        <TableInfo />
+        <TableData results={FakeBookings} />
```
dsobkow, line 11: This is a nice split into smaller components👍

You're passing props to this component in `Bookings.js` but not using it here and importing data from json file again. You should keep one or the other - pass data as prop to `SearchResults` and then pass it down to `TableData` or remove the prop and import from `Bookings.js` and keep this file as is.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+
+function TouristInfoCards() {
+  return (
+    <div className="container">
+      <div className="card">
```
dsobkow, line 6: The code responsible for creating a single card is repeated multiple times here. It's a sign that we should create a new `Card` component to reuse that code. My suggestion on how to tackle this would be to:
- create an array of objects that contain data for each city such as link, image source and summary
- map over that array and for each element create a Card component, passing data to that component with props

PR #134, opened 2020-07-13
OsmanElsahib-React-week-1

elsahib20:
Hello mentors,

Please have a look at my homework submission.

Thanks,
OsmanElsahib


src/App.js
```diff
@@ -2,12 +2,54 @@ import React from "react";
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <div className="album py-5">
+        <div className="container">
+          <div className="row">
+            <TouristInfoCards city={city1} />
+            <TouristInfoCards city={city2} />
+            <TouristInfoCards city={city3} />
```
dsobkow, line 47: I think that `TouristInfoCards` should be a container for cards that:
- gets data with information for each city (an array of objects would be more real-life scenario here)
- maps over that array and renders a `Card` component for each element of an array, passing data it needs to render as props

So you could either move all cities data to the `TouristInfoCards` component or extract it to a json file and import it in that component - the same way as `fakeBookings` data to make it even cleaner.

src/App.js
```diff
@@ -2,12 +2,54 @@ import React from "react";
...
+    city3 = {
+      name: "London",
+      link: "https://www.visitlondon.com/",
+      image:
+        "https://upload.wikimedia.org/wikipedia/commons/thumb/c/cd/London_Montage_L.jpg/275px-London_Montage_L.jpg",
+      text:
+        "London is the capital and largest city of England and the United Kingdom"
+    };
+  const footerInfo = [
+    "123 Fake Street, London, E1 4UD",
```
dsobkow, line 35: Nice idea to create a variable with footer info👍

src/Footer.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+function Footer(props) {
+  return (
+    <footer>
+      <ul>
+        {props.info.map((item, index) => {
+          return <li key={index}>{item}</li>;
```
dsobkow, line 8: Very good use of props and map! You even remembered about the key prop👏

src/Heading.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+function Heading() {
+  return (
+    <header className="App-header">
+      <img
+        className="App-logo"
+        src="https://image.flaticon.com/icons/svg/139/139899.svg"
```
dsobkow, line 8: The header looks good🙂

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+          <th scope="col">Room ID</th>
+          <th scope="col">Check-In Date</th>
+          <th scope="col">Check-Out Date</th>
+          <th scope="col">No. Nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map((item, index) => {
+          const nightsSpent = moment(item.checkOutDate).diff(
+            item.checkInDate,
```
dsobkow, line 22: Good use of the `moment` library here!

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+            "days"
+          );
+          return (
+            <tr key={index}>
+              <td scope="row">{item.id}</td>
```
dsobkow, line 27: Do you think that table header and body could be separate components?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+function TouristInfoCards(props) {
+  return (
+    <div className="col-md-4">
+      <div className="card mb-4">
```
dsobkow, line 6: As I mentioned in one of my previous comments, line 5 to 15 could be a `Card` component that is rendered here using map

src/App.js
elsahib: I did as you sugessted, I also found a couple of minor things that I've missed and got them corrected.

src/SearchResults.js
```diff
@@ -0,0 +1,50 @@
...
+              <td>{nightsSpent}</td>
+            </tr>
+          );
+        })}
+      </tbody>
+    </table>
+  );
+}
+
+function TableHead() {
```
dsobkow, line 33: Good job here, this is exactly what I meant. Although it would be cleaner to have each component in it's own file, so I would move `TableHead` component to `TableHead.js` file and import it here. Then you could create `TableRow` component and extract lines 15 to 25 there, passing the data it needs to render as props.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+function TouristInfoCards(props) {
```
dsobkow, line 3: You left 3 divs that belong to `TouristInfoCards` in the `App.js`, I think they should be inside this component and then you could map over `cities` inside those divs. I also think that lines 6 to 16 could be extracted to a new `Card` component that receives city info as props.

Well done with extracting cities data to json file, you could import it directly in this component - there is no need to pass it from the App as props.

src/SearchResults.js
elsahib: done.

src/TouristInfoCards.js
elsahib: and done.

PR #135, opened 2020-07-13
BuchraAtkeh React Homework Lesson1

Buchraatkeh1984:
please  review it when you are available

### DIFF
--- a/src/SearchResults.js
+++ b/src/SearchResults.js
@@ -1,5 +1,6 @@
 import React from "react";
 import Moment from "react-moment";
+import moment from "moment";
 
 const SearchResults = props => {
   return (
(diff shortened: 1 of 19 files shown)


src/SearchResults.js
```diff
@@ -0,0 +1,50 @@
...
+                <td>{result.email}</td>
+                <td>{result.roomId}</td>
+                <td>{result.checkInDate}</td>
+                <td>{result.checkOutDate}</td>
+                <td>
+                  {/* {moment(result.checkOutDate, "DD/MM/YYYY").diff(
+                  moment(result.checkInDate, "DD/MM/YYYY"),
+                  "days"
+                )} */}
+                  <Moment diff={result.checkInDate} unit="days">
```
dsobkow, line 38: I can see that you've used `react-moment` library, not `moment` that was specified in the instruction. I think it's fine as you used it correctly and it works well. So well done on figuring out how to use it🙂 However if you would decide to use `moment`, you were pretty close in the code that you commented out. This would work:
```
moment(result.checkOutDate).diff(result.checkInDate, "days")
```

PR #137, opened 2020-07-13
Nawal-React-lesson-1

NawalC:


src/App.js
```diff
@@ -1,13 +1,24 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        FooterArray={[
```
msachi, line 16: Try to stick to camel casing for props names, e.g. `footerArray`

src/Heading.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <div>
+      <header className="App-header">
+        {" "}
```
msachi, line 7: I think you don't need these, probably an autoformating issue?

src/SearchResults.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import moment from "moment";
+
+const SearchResults = props => {
+  return (
+    <div>
+      <table className="table">
+        <thead>
+          <tr>
```
msachi, line 9: You could extract this into a separate component, e.g. `HeadingRow`

src/SearchResults.js
```diff
@@ -0,0 +1,47 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {props.results.map((field, index) => {
+            const nightsSpent = moment(field.checkOutDate).diff(
+              field.checkInDate,
+              "days"
+            );
+            return (
+              <tr key={index}>
```
msachi, line 28: You could extract this into a separate component, e.g. `BodyRow`.

As a general rule, if you are mapping through something and returning a piece of JSX each time, then you can extract that JSX into a separate component.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+
+const TouristInfoCards = () => {
+  let TouristInfo = [
```
msachi, line 4: Camel casing for variable names also

src/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
...
+        "Find things to do in London, days out in London, London attractions and sightseeing and more"
+    }
+  ];
+
+  return (
+    <div className="card-container">
+      <div className="card-deck">
+        {TouristInfo.map((hotel, index) => {
+          return (
+            <div key={index} className="card">
```
msachi, line 33: You could extract this into a separate `Card` component

src/App.css
```diff
@@ -4,17 +4,20 @@
```
msachi, line 4: It's better to have a separate css file for each component

src/Search.js
```diff
@@ -1,23 +1,39 @@
...
+
+const Search = props => {
+  // console.log(props.search);
+  const [searchInput, setSearchInput] = useState("");
+  //const [onSubmit,setOnSubmit] = useState()
+
+  const handleSearchSubmit = event => {
+    event.preventDefault();
+    props.search(searchInput);
+  };
```
msachi, line 12: You could add `setSearchInput("")` at the end to clear the input when submitting the form (which is common practice)

src/Bookings.js
```diff
@@ -1,21 +1,52 @@
...
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState(null);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
+      //    fetch("https://cyf-react.glitch.me/delayed")
+      //fetch("https://cyf-react.glitch.me/error")
+      .then(response => response.json())
+      .then(data => {
+        if (data.error) {
+          throw data;
```
msachi, line 17: You probably want to throw just `data.error` here?

src/Bookings.js
```diff
@@ -1,21 +1,52 @@
...
+        }
+        console.log(data);
+        setBookings(data);
+        setLoading(false);
+      })
+      .catch(error => setError(error));
```
msachi, line 23: Nice of use of `catch` - this will also trigger e.g. if you get a generic error with a non-JSON response, and the JSON parsing fails.

msachi: Although in this case your `error.error` in the JSX below wouldn't work - one more reason to throw `data.error` instead of `data` above.

src/Bookings.js
```diff
@@ -1,21 +1,52 @@
...
+  }, []);
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    let filteredBooking = bookings.filter(
+      booking =>
+        booking.firstName.toLowerCase().includes(searchVal.toLowerCase()) ||
+        booking.surname.toLowerCase().includes(searchVal.toLowerCase())
+    );
+    setBookings(filteredBooking);
   };
```
msachi, line 34: Some spacing between functions would help with readability

src/Bookings.js
```diff
@@ -1,21 +1,52 @@
...
-
-  return (
-    <div className="App-content">
-      <div className="container">
-        <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+  if (loading && !error) {
```
msachi, line 35: You could make this slightly more readable with a ternary operator:
error ? (
...
) : isLoading ? (
...
) : (
...
)

src/BodyRow.js
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const BodyRow = props => {
+  const [isSelected, setIsSelected] = useState(null);
```
msachi, line 5: Nicely refactored 👍

PR #138, opened 2020-07-14
added lesson 1 exercises

SULAIMAN-100:


src/App.js
```diff
@@ -1,13 +1,23 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer
+        prop={[
```
lannem, line 15: You can call your props anything you like. In this case calling this prop something like `data` or `footerData` would be more descriptive than just calling it `prop`. That could lead to confusion if you have multiple props

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
...
+          <th scope="col">Email</th>
+          <th scope="col">Room ID</th>
+          <th scope="col">Check in date</th>
+          <th scope="col">Check out date</th>
+          <th scope="col">Nights</th>
+        </tr>
+      </thead>
+      {props.results.map(booking => {
+        return (
+          <tbody>
```
lannem, line 22: Having your `<tbody>` element inside the `return` statement of your `map` means that every row in the table has a `<tbody>` element wrapped around it. A `<table>` element should only have one `<tbody>`.

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
...
+        <tr>
+          <th scope="col">ID</th>
+          <th scope="col">Title</th>
+          <th scope="col">First Name</th>
+          <th scope="col">Surname</th>
+          <th scope="col">Email</th>
+          <th scope="col">Room ID</th>
+          <th scope="col">Check in date</th>
+          <th scope="col">Check out date</th>
+          <th scope="col">Nights</th>
```
lannem, line 17: This is good, however there is a way to do this that would save you having to have a line of code for each column. If you know every booking in the table will have data for all of these columns you could generate them dynamically using a `.map` function with `Object.keys`. If this doesn't make sense please let me know :)

src/Bookings.js
```diff
@@ -12,7 +13,7 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
```
lannem, line 16: This is a good example of using an appropriate prop name well done :)

PR #139, opened 2020-07-15
Laetitia-Week-1-React

supertata:


src/Heading.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <div>
+      <header className="App-header">
+        {" "}
```
lannem, line 7: There is no need for these: `{" "}`
It is just adding an empty string. The same goes for the ones on lines 12 and 13

src/Heading.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <div>
```
lannem, line 5: This `<div>` wrapper isn't really necessary as everything is wrapped inside the `<header>` tags :)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+
+const TouristInfoCards = () => {
+  let TouristInfo = [
```
lannem, line 4: The cards are good and work correctly, however it would be nice if the image, summary and destination matched the links!

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+const SearchResults = props => {
+  return (
+    <table className="table">
+      <thead>
+        <tr>
+          <th scope="">Id</th>
+          <th scope="col">Title</th>
+          <th scope="col">FirstName</th>
+          <th scope="col">Surname</th>
+          <th scope="">Email</th>
```
lannem, line 12: Not sure why the `col` scope is missing here and on ID?

src/Bookings.js
```diff
@@ -12,7 +13,8 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchButton />
```
lannem, line 16: I would include the SearchButton component inside the Search component rather than Bookings, as it is part of the functionality of the Search component

PR #140, opened 2020-07-15
HW Submission.

zision71:


src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
...
+      "https://images.unsplash.com/photo-1522961881734-7df1e2635d0c?ixlib=rb-1.2.1&auto=format&fit=crop&w=400&q=60",
+    summary:
+      "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s",
+    destination: "London"
+  }
+];
+function TouristInfoCards() {
+  return (
+    <div>
+      {data.map((i, index) => {
```
evanrolfe, line 32: Hint: you can make this code a bit simpler - you don't need the `{` and `return (` but can just put your jsx code right after the `=>` from the map:
```js
{data.map((i, index) =>
  <div key={index} className="card">
    <img src={i.image} className="card-img-top" />
    <div className="card-body">
      <p>{i.summary}</p>
      <a href={i.link} className="btn btn-primary">
        {i.destination}
      </a>
    </div>
  </div>
)}
```

src/components/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React, { useLayoutEffect } from "react";
+
+function Footer(props) {
+  return (
+    <ul>
+      {props.footerdata.map((i, index) => {
+        return <li key={index}>{i}</li>;
```
evanrolfe, line 7: This works, but another hint you can also do it like this:
```js
{props.footerdata.map((i, index) => <li key={index}>{i}</li>)}
```

src/components/SearchResults.js
```diff
@@ -0,0 +1,68 @@
...
+            <th scope="col">email</th>
+            <th scope="col">room id</th>
+            <th scope="col">check in date</th>
+            <th scope="col">check out date</th>
+            <th scope="col">No of nights</th>
+          </tr>
+        </thead>
+
+        <tbody>
+          {props.results.map((i, index) => {
```
evanrolfe, line 47: Its a little bit confusing that you use the variable name `i` here, because `i` is more commonly used to refer to `index`, so  although this works I would have used these variable names:
```js
{props.results.map((booking, i) => {
```

evanrolfe:
Really nice work Zee! You have just one issue with your step 1 solution:

You need to **only** extract the button from Search.js into its own component, specifically this button:
```html
<button className="btn btn-primary">Search</button>
```
So you will need to create a new component `components/SearchButton.js` and then import and use that component within the `components/Search.js`. You should not have to change App.js or any other files to complete step 1.

zision71:
> Really nice work Zee! You have just one issue with your step 1 solution:
> 
> You need to **only** extract the button from Search.js into its own component, specifically this button:
> 
> ```
> <button className="btn btn-primary">Search</button>
> ```
> 
> So you will need to create a new component `components/SearchButton.js` and then import and use that component within the `components/Search.js`. You should not have to change App.js or any other files to complete step 1.

Hi  Evan done necessary changes as you instruct. Regards.

evanrolfe:
> Hi Evan done necessary changes as you instruct. Regards.

Where can I see these changes? The code looks exactly the same as when I last looked at it?

zision71:
Zision-Lesson2 HW submission.

PR #142, opened 2020-07-16
completed exercises 1 to 7

Leidagandy:
Leida React week 1


src/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <ul>
+      {props.contactDetails.map((contact, index) => {
+        return <li key={index}>{contact}</li>;
```
lannem, line 7: The styling could potentially be nicer on your footer, for example the list items don't really need bullet points. This is me being picky though :)

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+          <tr>
+            <th scope="col">Id</th>
+            <th scope="col">Title</th>
+            <th scope="col">First Name</th>
+            <th scope="col">Last Name</th>
+            <th scope="col">Email</th>
+            <th scope="col">Room Id</th>
+            <th scope="col">Check in Date</th>
+            <th scope="col">Check out Date</th>
+            <th scope="col">No of Nigths</th>
```
lannem, line 18: You could potentially generate these <th> elements dynamically with a map function. It would save you some lines of code. Let me know if you're not sure what I mean!

src/TouristInfoCards.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+const TouristInfoCards = () => {
```
lannem, line 3: Rather than hard coding each card individually you could have stored the data in an Array of Objects and using a map function. This would save you lines of code and would make it easy to add additional cards in the future

Leidagandy:
Feedback implemented :)

PR #143, opened 2020-07-16
Solved React exercise of week-1

AltomHussain:


src/SearchButton.js
```diff
@@ -0,0 +1,16 @@
...
+function SearchButton() {
+  return (
+    <div className="search-row">
+      <input
+        type="text"
+        id="customerName"
+        className="form-control"
+        placeholder="Customer name"
+      />
+      <button className="btn btn-success">Search</button>
```
lannem, line 12: You have the right idea with this but the question was asking you to extract the `<button>` element only, so you only needed this line in your `return` statement

src/TouristInfoCards.js
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <div className="all-cards">
+      {cardInfo.map(info => {
+        return (
+          <div className="card">
+            <h2 className="city-title">{info.city}</h2>
+            <img src={info.image} className="card-img-top" alt="" />
+            <div className="card-body">
+              <p>{info.summary}</p>
+              <a href="{info.link}" className="btn btn-success">
```
lannem, line 43: The double quotes around `{info.link}` break the link so when you click on the button you do not get taken to the URL. Tiny mistake!

src/SearchResults.js
```diff
@@ -0,0 +1,44 @@
...
+          <th scope="col">nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map((data, index) => {
+          return (
+            <tr key={index}>
+              <th scope="row">{data.id}</th>
+              <td>{data.title}</td>
+              <td>d{data.firstName}</td>
```
lannem, line 27: You have a small error on this line where a `d` is placed before the first name on every row. Another tiny mistake :)

AltomHussain:
Thank you for reviewing my work appreciate it.

PR #144, opened 2020-07-16
completed 1 to 7 tasks

FarhanaFKhan:


src/App.js
```diff
@@ -1,13 +1,48 @@
 import React from "react";
-
+import Heading from "./Heading.js";
+import TouristInfoCards from "./TouristInfoCards.js";
 import Bookings from "./Bookings";
+import Footer from "./Footer.js";
 import "./App.css";
 
+const cities = [
```
Dedekind561, line 8: Hi @FarhanaFKhan, loved that you've abstracted the cities data into an array here.

src/App.js
```diff
@@ -1,13 +1,48 @@
...
+  {
+    name: "Manchester",
+    website: "visitmanchester.com",
```
Dedekind561, line 11:
```suggestion
    website: "https://visitmanchester.com",
```

Dedekind561: Without the full url here then the link doesn't take me to the right site - in fact, it just links me back to the same page.

src/App.js
```diff
@@ -1,13 +1,48 @@
...
+  {
+    name: "London",
+    website: "visitlondon.com",
+    image:
+      "https://cdn.londonandpartners.com/-/media/images/london/visit/general-london/towerbridgecopyrightvisitlondoncomantoinebuchet.jpg?mw=1920&hash=B6672601055B38B34C592E0160B0445AA063B876",
+    description:
+      "Welcome to Visit London, your official city guide to London, England. Find things to do in London, days out in London, London attractions and sightseeing, what's on, London events, theatre, tours, restaurants and hotels in London. Plan your trip to London with useful traveller information"
+  }
+];
+const footerArr = [
```
Dedekind561, line 34: 😄 - Good use of props for the footer

src/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
```
Dedekind561, line 4:
```suggestion
import fakeBookings from "./data/fakeBookings.json";
```
Keep this variable starting with a lowercase - I would only capitalise variables if they represent a React component

src/Footer.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+
+function Footer(props) {
```
Dedekind561, line 3: Consider destructuring `footerArr` here 👍

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import moment from "moment";
+moment().format();
```
Dedekind561, line 3: Not sure what this line is doing at the moment (line 3) - did you leave this in by accident ?

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+          <th scope="col">Email</th>
+          <th scope="col">Room id</th>
+          <th scope="col">Check In Date</th>
+          <th scope="col">Check Out Date</th>
+          <th scope="col">Number of Nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map(result => {
+          let arrivalDate = moment(result.checkInDate, "YYYY-MM-DD");
```
Dedekind561, line 23:
```suggestion
          const arrivalDate = moment(result.checkInDate, "YYYY-MM-DD");
```

Dedekind561: I would consider defaulting to use the `const` keyword here as you don't intend on re-assigning these variables.
But it isn't a huge deal 😄

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+          let departureDate = moment(result.checkOutDate, "YYYY-MM-DD");
+          let numberOfNights = departureDate.diff(arrivalDate, "days");
+          return (
+            <tr key={result.id}>
+              <th scope="row">{result.id}</th>
+              <td>{result.title}</td>
+              <td>{result.firstName}</td>
+              <td>{result.surname}</td>
+              <td>{result.email}</td>
+              <td>{result.roomid}</td>
```
Dedekind561, line 33:
```suggestion
              <td>{result.roomId}</td>
```
The `result` object has a `roomId` property so this isn't displaying in the browser

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+          <th scope="col">Room id</th>
+          <th scope="col">Check In Date</th>
+          <th scope="col">Check Out Date</th>
+          <th scope="col">Number of Nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map(result => {
+          let arrivalDate = moment(result.checkInDate, "YYYY-MM-DD");
+          let departureDate = moment(result.checkOutDate, "YYYY-MM-DD");
```
Dedekind561, line 24: Good clear variable names here, well done! 🎉

src/TouristInfoCards.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import CityCard from "./CityCard.js";
+
+function TouristInfoCards(props) {
+  const card = (
+    <div className="Tourist-Cards">
+      {props.cities.map(city => {
+        return <CityCard key={city.name} city={city} />;
```
Dedekind561, line 8: Love that you've abstracted the city card stuff into a separate component `CityCard`

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+          let numberOfNights = departureDate.diff(arrivalDate, "days");
+          return (
+            <tr key={result.id}>
```
Dedekind561, line 27: Could consider abstracting some of this component into a smaller react component like `BookingEntry` or something like that - in charge of displaying a single row in the bookings table

src/Order.js
```diff
@@ -0,0 +1,16 @@
...
+import RestaurantButton from "./RestaurantButton";
+
+function Order({ orderType }) {
+  const [orders, setOrders] = useState(0);
+  function orderOne() {
+    setOrders(orders + 1);
+  }
+  return (
+    <li className="order-list">
+      {orderType}: {orders} <RestaurantButton eventHandler={orderOne} />
```
Dedekind561, line 11: I'd avoid calling it `eventHandler` as that is a very general name for a function  - indeed it is true of any function that gets called when an event occurs. Instead I'd think about using a more descriptive name relevant to your app and what it does to the `orders`.

src/Order.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order({ orderType }) {
+  const [orders, setOrders] = useState(0);
```
Dedekind561, line 5: Good naming conventions used on line 5 😄

src/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
   return (
-    <div>
+    <div className="restaurant">
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+      <ul className="ul-styling">
+        <Order orderType="Pizzas" />
```
Dedekind561, line 9: Good stuff here @FarhanaFKhan - think of another way you could also render each of the orders without rendering them each individually one at a time.

src/RestaurantButton.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+function RestaurantButton({ eventHandler }) {
```
Dedekind561, line 3: Again same as before - think of a different name for this function you're destructuring

src/TableRow.js
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+function TableRow({ result }) {
+  const [highlight, setHighlight] = useState(false);
```
Dedekind561, line 5: So you've got the right idea but it is useful to be aware of naming conventions for booleans in this section. If you have a variable that stores a boolean value like a number you could think of a name that starts with a `is` or `has` e.g. `const isEven = true;`. The variable name `isEven` sort of reads like a question

Dedekind561: So think about applying the above when you create the state variable on line 5 - what does this variable reflect about your UI ?

src/SearchResults.js
```diff
@@ -0,0 +1,30 @@
...
+          <th scope="col">Email</th>
+          <th scope="col">Room id</th>
+          <th scope="col">Check In Date</th>
+          <th scope="col">Check Out Date</th>
+          <th scope="col">Number of Nights</th>
+        </tr>
+      </thead>
+      <tbody>
+        {props.results.map(result => {
+          return <TableRow key={result.id} result={result} />;
```
Dedekind561, line 22: Good this section looks really clear now when you `map` over the resutls and display a single TableRow 😄

src/SearchResults.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import TableRow from "./TableRow.js";
+
+function SearchResults(props) {
```
Dedekind561, line 4: Another chance for a bit of destructuring on line 4 👍

src/TableRow.js
```diff
@@ -0,0 +1,32 @@
...
+
+  function highlightRow() {
+    setHighlight(!highlight);
```
Dedekind561, line 8: 👍

src/Bookings.js
```diff
@@ -1,18 +1,24 @@
...
+import SearchResults from "./SearchResults.js";
 
 const Bookings = () => {
+  const [bookings, setBookings] = useState([]);
   const search = searchVal => {
     console.info("TO DO!", searchVal);
   };
 
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
```
Dedekind561, line 12: Where you've done this logic for making the network request you could think about abstracting this code into a separate function called something like `fetchBookings` and then invoke it inside of `useEffect`

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+  const [newBookingInfo, setNewbookingInfo] = useState({
+    id: "0",
+    title: "",
+    firstName: "",
+    surname: "",
+    email: "",
+    roomId: "",
+    checkInDate: "",
+    checkOutDate: "",
+    items: []
```
Dedekind561, line 15: This `items` array looks like you're trying to reference the whole array of bookings - instead you can just deal with the whole array of bookings in the parent component.

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
+import React, { useState } from "react";
+import SearchResults from "./SearchResults.js";
+import TableRow from "./TableRow.js";
+function BookingForm({ handleNewBooking }) {
```
Dedekind561, line 4: Good use of destructuring here

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+    email: "",
+    roomId: "",
+    checkInDate: "",
+    checkOutDate: "",
+    items: []
+  });
+
+  function handleSelect(event) {
+    let newId = 5;
+    newId++;
```
Dedekind561, line 20: `newId` will always be fixed as 6 as the value will be reset on every call and doesn't persist in state

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+    const updatedCustomerInfo = {
+      id: newId,
+      title: event.target.value
+    };
+    setNewbookingInfo(updatedCustomerInfo);
+  }
+
+  function handleChange(event) {
+    const updatedCustomerInfo = {
```
Dedekind561, line 29: This looks good - updating the state of the controlled component only -namely, the fields of the form

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+      ...newBookingInfo,
+      [event.target.name]: event.target.value
+    };
+    setNewbookingInfo(updatedCustomerInfo);
+  }
+
+  function handleSubmit(event) {
+    event.preventDefault();
```
Dedekind561, line 37: 👍

Dedekind561: For `event.preventDefault()`

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+    let items = [];
```
Dedekind561, line 38: Not sure why you need this `items` array - you can update the array of bookings at the parent level. The only thing you will need to be concerned with when the submit event fires is passing the new booking from the form up to the parent component ( `App.js`, I believe )

src/BookingForm.js
```diff
@@ -0,0 +1,195 @@
...
+      id: "0",
+      title: "",
+      firstName: "",
+      surname: "",
+      email: "",
+      roomId: "",
+      checkInDate: "",
+      checkOutDate: ""
+    });
+    handleNewBooking(newBookingInfo.items);
```
Dedekind561, line 62: So when this function is invoked it only needs to be passed your new booking - handle the updating array of bookings logic in the component where the bookings state was initialised

src/BookingForm.js
Dedekind561: DM me if you need some help fixing this one

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from "react";
+
+function CustomerProfile({ id }) {
+  const [customerInfo, setCustomerInfo] = useState({});
+  useEffect(() => {
+    if (!id == 0) {
```
Dedekind561, line 6: It looks to me like condition `!id == 0` this will always evalute to true when `id` is truthy. Not sure what this condition is checking - remember that React will only trigger this effect if the `id` in the dependency array on line 11 updates.

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
...
+      fetch(`https://cyf-react.glitch.me/customers/${id}`)
+        .then(res => res.json())
+        .then(data => setCustomerInfo(data));
+    }
+  }, [id]);
+
+  if (id != 0) {
```
Dedekind561, line 13: Why are you guarding against the `id` being `0` here ?

src/SearchResults.js
```diff
@@ -0,0 +1,58 @@
+import React, { useState, useEffect } from "react";
+import TableRow from "./TableRow.js";
+import CustomerProfile from "./CustomerProfile.js";
+
+function SearchResults(props) {
+  const [profileId, setProfileId] = useState(0);
+  const [rows, setRows] = useState(props.results);
```
Dedekind561, line 7: On line 7 you are initalising some state using _props_ from the parent component. This is an anti-pattern in React, indeed it means that the component won't update the state in the way you expect. If in doubt  remember that the props are read-only and should change when they are passed through to the component from above.

src/Search.js
```diff
@@ -1,23 +1,35 @@
...
 
-const Search = () => {
+const Search = ({ search }) => {
+  const [searchInput, setSearchInput] = useState("");
+  function handleOnChange(event) {
+    console.log(event.target.value);
+    setSearchInput(event.target.value);
+  }
+  function handleSubmit(event) {
+    search(searchInput);
```
Dedekind561, line 11: This all looks good to me :)

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+        setBookings(data);
+        setLoading(true);
+      })
+      .catch(error => window.alert("Ooops, Something went wrong"));
+  }, []);
 
-const Bookings = () => {
   const search = searchVal => {
-    console.info("TO DO!", searchVal);
+    let searchRes = bookings.filter(booking => {
```
Dedekind561, line 20: This looks good to me too , well done 👍 . Just remember that the update on the UI isn't happening because you're saving props in state in one of the child components.

Dedekind561, line 20:
```suggestion
    let filteredBookings = bookings.filter(booking => {
```

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+      return (
+        booking.firstName.toLowerCase().includes(searchVal) ||
```
Dedekind561, line 22: Good use of `filter` here 😄

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+const Bookings = ({ newBooking }) => {
+  const [bookings, setBookings] = useState([]);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => {
+        setBookings(data);
+        setLoading(true);
```
Dedekind561, line 14: Here you are effectively saying that the `loading` state is `true` _after_ the bookings have been fetched from the API. It should be that the component has stopped `loading` once the `fetch` has been successful - you may need to think about how the `loading` is intialised

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+
+const Bookings = ({ newBooking }) => {
+  const [bookings, setBookings] = useState([]);
+  const [loading, setLoading] = useState(false);
```
Dedekind561, line 7:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+    });
+    setBookings(searchRes);
   };
 
   return (
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        {loading ? (
```
Dedekind561, line 33: This currently reads as `loading === true` -> render the search results
`loading === false` -> render the loading symbol

src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+          <SearchResults results={bookings} newBookingInfo={newBooking} />
+        ) : (
+          <h1>Loading...</h1>
```
Dedekind561, line 36:
```suggestion
          <p>Loading...</p>
```

Dedekind561: This is super super picky but shouldn't be a `h1` element as a screen reader or some other accessibility tool would make this appear way more important than it is on the page.

PR #151, opened 2020-07-19
BuchraAtkeh React Homework Lesson2

Buchraatkeh1984:
please review my homework


src/Restaurant.js
```diff
@@ -1,14 +1,16 @@
...
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <OrderPizza />
+        <OrderSalad />
+        <OrderChocolatCake />
```
dsobkow, line 13: The idea here was to have one reusable `Order` component and pass `orderType` as props, so it should look something like this:
```
  <ul>
    <Order orderType="Pizzas" />
    <Order orderType="Salads" />
    <Order orderType="Chocolate cake" />
  </ul>
```

src/RestaurantButton.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+const RestaurantButton = ({ handler, message }) => {
+  return (
+    <button onClick={handler} className="btn btn-primary restBut">
```
dsobkow, line 4: Well done with extracting the button and giving it an `onClick` handler🙂 I think it would make more sense to call the prop with the actual function name - `orderOne` instead of `handler`. It's more intuitive if the function name starts with a verb. Also I don't think that passing message is necessary as you already have order type in your Order component.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div className="cardsContainer">
+      {props.cities.map(city => {
+        return (
+          <div className="card">
+            <img
```
dsobkow, line 9: I think I have suggested extracting lines 8 to 21 to a separate `Card` component in my previous review?🙂

src/OrderPizza.js
```diff
@@ -0,0 +1,65 @@
...
+//           function orderOne(orderType){
+//       if (orderType=="Pizza")
+//         setOrderP(orderP + 1);
+//         else if (orderType=="Salad")
+//         setOrderS(orderS + 1);
+//         else
+//         setOrderC(orderC + 1);
+//           }
+//     return(
+//       // orderType.map(order =>{
```
dsobkow, line 22: Lots of commented out code - always delete it before committing

src/OrderChocolatCake.js
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from "react";
+
+import RestaurantButton from "./RestaurantButton";
+
+const OrderChocolateCake = () => {
+  //const types=["Pizza","Salad","Chocolate cake"]
+  const [order, setOrder] = useState(0);
+
+  function orderOne() {
```
dsobkow, line 9: State and `onClick` function work very well together - well done! But please refactor your code so there is only one reusable `Order` component (not 3 different components) and make use of props🙂 If you pass `orderType` prop to this component, you will be able to replace hard coded ChocolateCake on line 17 with that prop.

And please delete all comments🙂

src/SearchResults.js
```diff
@@ -0,0 +1,78 @@
...
+  //     isClicked=false;
+  //     e.target.classList.remove("highLight")
+  //   }
+  // }
+  return (
+    <div className="table-responsive-sm">
+      <table className="table table-bordered">
+        <thead className="table-primary fontColor">
+          <tr
+            onClick={e => {
```
dsobkow, line 11: Table header should not have `onClick` event. It should be only applied to each table row inside table body.

src/SearchResults.js
```diff
@@ -0,0 +1,78 @@
...
+            <th scope="col">Duration of stay </th>
+          </tr>
+        </thead>
+        <tbody>
+          {props.results.map(result => {
+            return (
+              <tr
+                onClick={e => {
+                  if (e.currentTarget.classList.contains("highLight")) {
+                    e.currentTarget.classList.remove("highLight");
```
dsobkow, line 37: That's clever solution🙂 However it's not the way it should be done in React. I would suggest following these steps:
1. Extract code from lines 47 to 70 to a new `TableRow` component - you can have a look at the solution to lesson 1 exercises and do the same.
2. Inside that new `TableRow` component declare a new state `isSelected` with the corresponding setter function `setIsSelected` and initially set the value to `false`.
3. Pass a function to the `onClick` handler that changes `isSelected` to the opposite value when clicked:
```
() => setIsSelected(!isSelected)
```
`!` operator will basically change `true` to `false` and `false` to `true`.
4. Then you can add a `className` attribute to your `<tr>` element and within a curly braces add a ternary operator that assigns a class name depending on the value of `isSelected`. For example if it's true - assign the class name with highlighted background - otherwise assign a different one or even null:
```
className={isSelected ? "highLight" : null}
```

PR #152, opened 2020-07-19
Nawal-CYF-hotel-Lesson2

NawalC:


src/Order.js
```diff
@@ -0,0 +1,15 @@
...
+import RestaurantButton from "./RestaurantButton";
+
+const Order = ({ orderType }) => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
+    setOrders(orders + 1);
+  };
+  return (
+    <li>
+      {orderType}: {orders} <RestaurantButton totalOrders={orderOne} />
```
msachi, line 11: I would call this prop `orderOne`, would be more intuitive than `totalOrders`. It also indicates that it's a function.

src/Bookings.js
```diff
@@ -1,21 +1,32 @@
...
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+//import FakeBookings from "./data/fakeBookings.json";
+import SearchButton from "./SearchButton.js";
 
 const Bookings = () => {
+  const [bookings, setBookings] = useState(null);
```
msachi, line 8: Not strictly necessary but usually it's better to initialise with the same data type, in this case an array.

If you change it make sure that your `if` statement on line 19 is also adjusted.

src/SearchResults.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from "react";
+import moment from "moment";
+const SearchResults = props => {
+  const [isSelected, setIsSelected] = useState([]);
```
msachi, line 4: Clever approach and totally works. If you want to simplify it, try the steps I sent you on slack.

PR #153, opened 2020-07-20
OsmanElsahib-React-week2

elsahib:
Hello Mentors,

Please have a look at and review my submission for this weeks homework.

kind regards,


src/Components/Bookings.js
```diff
@@ -0,0 +1,28 @@
+import React, { useState, useEffect } from "react";
+import Search from "./Search.js";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "../data/fakeBookings.json";
+
+const Bookings = () => {
+  const search = searchVal => {
+    console.info("TO DO!", searchVal);
+  };
+  const [bookings, setBookings] = useState([]);
```
dsobkow, line 10: Great use of React hooks and fetch method!👏 As you're fetching your data now, there is no need to import it on line 4 any more. Also the convention is to keep the state at the top of the component, so I would move it to line 7

src/Components/Card.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+function Card(props) {
+  return (
```
dsobkow, line 4: Very nice! Maybe you could destructure props here?🙂

src/Components/Order.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = ({ orderType }) => {
+  const [orders, setOrders] = useState(0);
+  const OrderOne = () => {
```
dsobkow, line 6: State and `onClick` function work very well together - well done! This is a really nice and clean component👏 
 I'm just missing some blank lines for better readability - after `useState` and after `orderOne` function🙂 also the function name should start with lower case

src/Components/Restaurant.js
```diff
@@ -0,0 +1,17 @@
...
+import Order from "./Order";
+
+const Restaurant = () => {
+  return (
+    <div className="container">
+      <h3>Restaurant Orders</h3>
+      <ul>
+        <Order orderType="Pizzas" />
+        <Order orderType="Salads" />
+        <Order orderType="Chocolate cake" />
```
dsobkow, line 11: Well done on reusing the `Order` component! ✅

src/Components/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const RestaurantButton = ({ clickAction }) => {
+  return (
+    <button onClick={clickAction} className="btn btn-primary">
```
dsobkow, line 5: Nice one! I think it would be more intuitive to call the prop with the actual function name - `orderOne` instead of `clickAction`.

src/Components/TouristInfoCards.js
```diff
@@ -0,0 +1,19 @@
...
+import React from "react";
+import Card from "./Card.js";
+
+function TouristInfoCards() {
+  return (
+    <div className="album py-5">
+      <div className="container">
+        <div className="row">
+          {Cities.map((city, index) => {
+            return <Card city={city} key={index} />;
```
dsobkow, line 11: Excellent! Now we have clear separation of concerns here

src/Components/TableRow.js
```diff
@@ -0,0 +1,26 @@
+import React, { useState } from "react";
+
+function TableRow(props) {
+  const [selectedRow, setSelectedRow] = useState(null);
+  const setRowStatus = () => {
+    if (selectedRow === null) {
+      setSelectedRow("table-active");
+    } else {
+      setSelectedRow(null);
```
dsobkow, line 9: That's a clever solution🙂 However I would suggest slightly different approach:
1. The value in state could be a boolean called `isSelected` initially set to `false`.
3. The `onClick` handler could be a function that changes `isSelected` to the opposite value when clicked:
```
setIsSelected(!isSelected)
```
`!` operator will basically change `true` to `false` and `false` to `true`.
4. Then you could add a ternary operator when defining a `className` attribute, that assigns a class name depending on the value of `isSelected`. For example if it's true - assign the class name with highlighted background - otherwise assign a different one or even null:
```
className={isSelected ? "table-active" : null}
```

elsahib: I started off with this approach actually but got stuck then came up with this solution. I got stuck on where should I have my ternary statement, but now that you handed it to me I'll re-werite it.

src/Components/Card.js
elsahib: done

src/Components/Bookings.js
elsahib: done

src/Components/RestaurantButton.js
elsahib: done

src/Components/Order.js
elsahib: done

src/Components/TableRow.js
elsahib: Done

elsahib:
This was very interesting exposure to how intuitive is coding using react.

PR #154, opened 2020-07-21
lesson  2 submitted

zision71:
lesson  2 submitted


evanrolfe:
@zision71 all the points are fine except for the last one (16. Highlight booking row when clicked) - when I click a row, it highlights all of the rows but really it should only highlight the row that was clicked. Please fix that and I'll have a look. Thanks

PR #158, opened 2020-07-23
Lesson 2 compelet

Hedyeh65:


src/components/CreateCards.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+function CreateCards(props) {
+  return (
+    <div className="container">
+      {props.cardInfo.map((data, index) => {
+        return (
```
dsobkow, line 7: I think that this component should render a single card, so div with `container` class from line 5 and map function should stay in `TouristInfoCards` component. Also React component names should be nouns - I wouldn't use word `create` in the name as it's a verb. And I would stick to arrow functions🙂

src/components/Order.js
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order(props) {
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    return setOrders(orders + 1);
+  }
```
dsobkow, line 9: State and `onClick` function work very well together - well done! This is a really nice and clean component👏

src/components/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order orderType={"Pizzas"} />
+        <Order orderType={"Salad"} />
+        <Order orderType={"Choclate Cake"} />
```
dsobkow, line 11: Well done on reusing the `Order` component! ✅

src/components/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function RestaurantButton(props) {
+  return (
+    <button className="btn btn-primary" onClick={props.customerOrders}>
+      Add
```
dsobkow, line 6: Nice one! I think it would be more intuitive to call the prop with the actual function name - `orderOne` instead of `customerOrders`.

src/components/SearchResults.js
```diff
@@ -0,0 +1,22 @@
...
+import TableInfo from "./TableInfo.js";
+import TableData from "./TableData.js";
+
+function SearchResults() {
+  const [bookings, setBookings] = useState([]);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me.")
+      .then(response => response.json())
+      .then(data => setBookings(data));
+  }, []);
```
dsobkow, line 11: Great use of React hooks and fetch method!👏 But according to instructions, data should be fetched in the `Bookings` component and passed to `SearchResults` as props.

src/components/TableData.js
```diff
@@ -0,0 +1,13 @@
+import React, { useState } from "react";
+
+import TableRow from "./TableRow";
+function TableData(props) {
+  return (
+    <tbody>
+      {props.results.map((data, index) => {
+        return <TableRow key={index} data={data} />;
+      })}
```
dsobkow, line 9: Since you have a new `TableRow` component, I think there is no point to have `TableData` component. I would move this code to `SearchResults` so the props are not passed through too many levels.

src/components/TableRow.js
```diff
@@ -0,0 +1,26 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+function TableRow(props) {
+  const [isClicked, setIsClicked] = useState(true);
+  return (
+    <tr
+      onClick={() => setIsClicked(!isClicked)}
+      style={!isClicked ? { backgroundColor: "yellow" } : null}
```
dsobkow, line 9: Very nice solution, well done!👏 I think it would make more sense if `isClicked` was initially set to false. And if `isClicked` is true, then the background color should change.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import CreateCards from "./CreateCards";
+
+const Information = [
+  {
+    link: "https://peoplemakeglasgow.com/",
+    image:
```
dsobkow, line 7: the `Information` array contains a lot of static data that will not change, so I think the component would be cleaner if this data was moved to a json file and just imported here.

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,33 @@
...
+      "https://cdn.londonandpartners.com/visit/general-london/areas/river/76709-640x360-houses-of-parliament-and-london-eye-on-thames-from-above-640.jpg",
+
+    summary:
+      " London is the capital and largest city of England and the United Kingdom. Standing on the River Thames in the south-east of England, at the head of its 50-mile (80 km) estuary leading to the North Sea,London has been a major settlement for two millennia. Londinium was founded by the Romans."
+  }
+];
+
+function TouristInfoCards() {
+  return <CreateCards cardInfo={Information} />;
+}
```
dsobkow, line 31: Following my comment for `CreateCards` component, `TouristInfoCards` should look something like this:
```
const TouristInfoCards = () => (
  <div className="container">
    {Information.map((data, index) => {
      return <Card key={index} cardInfo={data} />;
    })}
  </div>
);
```

PR #159, opened 2020-07-23
lesseon 2 done

Iman-feisali:
lesson 2 and some changes in lesson 1


src/Restaurant.js
```diff
@@ -1,14 +1,15 @@
 import React from "react";
 
+import Orders from "./Order";
```
Dedekind561, line 3:
```suggestion
import Order from "./Order";
```
You need to double check this one - you're importing a singular `Order` here :)

src/Restaurant.js
```diff
@@ -1,14 +1,15 @@
...
 const Restaurant = () => {
-  const pizzas = 0;
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Orders orderType={"Pizzas"} />
```
Dedekind561, line 10: 👍 - this looks good :) How could you render each of the Orders without rendering the Order component 3 times in a row. Think of how you do this with other components.

src/RestaurantButton.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const RestaurantButton = props => {
```
Dedekind561, line 3: 👍

src/SearchResults.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState, useEffect } from "react";
+import TableHeader from "./TableHeader";
+import TableData from "./TableData";
+
+function SearchResults() {
+  useEffect(() => {
```
Dedekind561, line 6: This looks great to me :)

src/SearchResults.js
```diff
@@ -0,0 +1,27 @@
...
+    console.log("hello react");
+    fetch("https://cyf-react.glitch.me")
+      .then(res => {
+        return res.json();
+      })
+      .then(data => {
+        return setBookings(data);
+      });
+  }, []);
+  const [bookings, setBookings] = useState([]);
```
Dedekind561, line 16: Makes a little more sense to declare your state and the setter function at the top of your functional component - as you will need them later on

src/TableData.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import TableRow from "./TableRow";
+
+function TableData(props) {
+  return (
+    <tbody>
+      {props.results.map((item, index) => {
+        return <TableRow item={item} key={index} />;
```
Dedekind561, line 8: Good naming here in the map 😄

src/TableHeader.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+
+function TableHeader() {
+  return (
+    <thead>
+      <tr>
+        <th scope="col">#</th>
```
Dedekind561, line 7: Again here you could think about another way of rendering each of the columns without repeatedly writing out the `<th>...</th>` each time

src/TableRow.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <tr
+      id="table-row"
+      className={handleClick ? "tr-selected" : null}
+      onClick={() => {
+        setHandleClick(!handleClick);
+      }}
+    >
+      <th scope="row" />
+      <td>{props.item.id}</td>
```
Dedekind561, line 15: Same as before with the rendering of individual `td` elements

src/TableRow.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const TableRow = props => {
```
Dedekind561, line 4: Consider destructuring in here to access the `item` variable

src/TableRow.js
```diff
@@ -0,0 +1,30 @@
...
+  const [handleClick, setHandleClick] = useState(false);
```
Dedekind561, line 5: `handleClick` sounds like the name of a function and not the name of a *boolean*

Dedekind561: Hint: a good name for a variable would be something that starts with `is` or `has` something like that
Also think about what this state value represents

Iman-feisali:
reopen for update

PR #160, opened 2020-07-24
Shahd-React-week2

shahdalhaj:


src/Bookings.js
```diff
@@ -1,21 +1,33 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+//import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults";
 
 const Bookings = () => {
+  const [Bookings, setBookings] = useState([]);
```
msachi, line 7: Try to use camelCasing, i.e. `bookings`

src/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <div className="footer">
+      <ul>
+        {props.info.map(function(item, index) {
```
msachi, line 7: Try to stick to one way of declaring functions, e.g. arrow functions (usually the cleanest way inside JSX)

src/Restaurant.js
```diff
@@ -1,14 +1,13 @@
 import React from "react";
+import Order from "./Order";
 
 const Restaurant = () => {
-  const pizzas = 0;
   return (
     <div>
```
msachi, line 6: As a general rule use only as much HTML as you need, so unless it's for styling purposes, no need to wrap things in divs.

src/RestaurantButton.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const RestaurantButton = ({ clickEvent }) => {
+  return (
+    <div>
```
msachi, line 5: Same here...

src/TableDetails.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const TableDetails = props => {
+  const [isSelected, setIsSelected] = useState([]);
```
msachi, line 5: This way works, but there is an easier (cleaner) way that doesn't involve handling ids.

You can extract each table row into a separate component, then add state to each row component to keep track of whether it's selected or not (it would just be a boolean). Finally you need to make className dependent on the value in the state.

Let me know if you want to give it a try and need help with this.

PR #161, opened 2020-07-24
Laetitia-React-Week 2

supertata:


src/components/Bookings.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useEffect } from "react";
+import "./Bookings.css";
+import Search from "./Search.js";
+import SearchResults from "./SearchResults.js";
+import SearchButton from "./SearchButton.js";
+
+const Bookings = () => {
+  const [bookings, setBookings] = useState(null);
```
lannem, line 8: This should be `useState([])` instead of `useState(null)`

src/components/Bookings.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useEffect } from "react";
+import "./Bookings.css";
+import Search from "./Search.js";
+import SearchResults from "./SearchResults.js";
+import SearchButton from "./SearchButton.js";
```
lannem, line 5: This import isn't used so should be removed otherwise you will get linting warnings

src/components/Bookings.js
```diff
@@ -0,0 +1,35 @@
...
+
+const Bookings = () => {
+  const [bookings, setBookings] = useState(null);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
+      .then(response => response.json())
+      .then(data => {
+        console.log(data);
```
lannem, line 13: When you are coding in production for a company you should generally remove any `console.log` statements you used fro debugging, nothing wrong with it in here though

src/components/Bookings.js
```diff
@@ -0,0 +1,35 @@
...
+      <div className="App-content">
+        <div className="container">
+          <Search search={search} />
+
+          <SearchResults results={bookings} />
+        </div>
+      </div>
+    );
+  } else {
+    return <div>Loading</div>;
```
lannem, line 31: Nice touch with the Loading div 👍

src/components/SearchResults.js
```diff
@@ -0,0 +1,64 @@
+import React, { useState } from "react";
+import "./SearchResults.css";
+import moment from "moment";
+const SearchResults = props => {
+  const [isSelected, setIsSelected] = useState([]);
+  const setClassName = id => {
```
lannem, line 6: Well done on getting this function to work, the logic is good! However, you could make it a bit simpler by using a boolean check for `isSelected` rather than a String

PR #162, opened 2020-07-24
React -Week -2

Taslima26:
Sorry for submitting after solution .I forgot to do it.
Thank you

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,26 @@
 import React from "react";
 
 import Bookings from "./Bookings";
+import Header from "./Heading";
+import DisplayTouristCard from "./DisplayTouristCard";
 import "./App.css";
+import "./TouristInfoCard.css";
+import Footer from "./Footer";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Header />
+      <DisplayTouristCard />
       <Bookings />
+
+      <Footer
+        footerData={[
+          "123 Fake Street, London, E1 4UD",
+          "hello@fakehotel.com",
+          "0123 456789"
+        ]}
+      />
     </div>
   );
 };
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 
 import Bookings from "./Bookings";
 import Header from "./Heading";
@@ -6,6 +6,8 @@ import DisplayTouristCard from "./DisplayTouristCard";
 import "./App.css";
 import "./TouristInfoCard.css";
 import Footer from "./Footer";
+import Restaurant from "./Restaurant";
+import RestaurantButton from "./RestaurantButton";
 
 const App = () => {
   return (
@@ -13,6 +15,8 @@ const App = () => {
       <Header />
       <DisplayTouristCard />
       <Bookings />
+      <Restaurant />
+      <RestaurantButton />
 
       <Footer
         footerData={[
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,7 +1,9 @@
 import React from "react";
 import Search from "./Search.js";
+import SearchResults from "./SearchResults.jsx";
+
 // import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
   const search = searchVal => {
@@ -11,9 +13,9 @@ const Bookings = () => {
   return (
     <div className="App-content">
       <div className="container">
-        <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
       </div>
+      <Search search={search} />
     </div>
   );
 };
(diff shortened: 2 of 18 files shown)


src/App.js
```diff
@@ -1,13 +1,30 @@
-import React from "react";
+import React, { useState } from "react";
```
timhamrouge, line 1:
```
import React from "react";
```

Remember to remove your unused imports after the code that uses them moves to different components :)

src/Bookings.js
```diff
@@ -1,19 +1,36 @@
...
+
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
+import SearchResults from "./SearchResults.jsx";
+
 // import SearchResults from "./SearchResults.js";
 // import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
+  const [BookingsData, setBokings] = useState([]);
```
timhamrouge, line 17:
```
```

👍 well done with the useState hook! Be careful of little typos though, and the exercise was asking for a "bookings" variable. But it all works, so good stuff!

src/Bookings.js
```diff
@@ -1,19 +1,36 @@
...
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`)
+      .then(res => res.json())
+      .then(data => setBokings(data));
+  }, []);
+  console.log(BookingsData);
```
timhamrouge, line 23: Well done with the useEffect hook! The console.log() is meant to be in scope of the useEffect hook though, where could we move it to?

src/Order.jsx
```diff
@@ -0,0 +1,27 @@
...
+import "./Order";
+const Order = props => {
+  const [foodItems, countFoodItems] = useState(0);
+  const incrementCounter = () => {
+    countFoodItems(foodItems + 1);
+  };
+  return (
+    <div>
+      {props.orderType}:{foodItems}
+      <button style={{ marginLeft: "30px" }} onClick={incrementCounter}>
```
timhamrouge, line 20: 👍  - Good work with the functionality incrementing the counters here, and using the useState hook. You actually have a Component declared that is intended to take the place of this button... see below...

src/Restaurant.js
```diff
@@ -1,13 +1,29 @@
-import React from "react";
-
+import React, { useState } from "react";
```
timhamrouge, line 1:
```
import React from "react";
```

The useState import has moved to a different file again here. This isn't a massive problem, but you will see warning in your console!

src/Restaurant.js
```diff
@@ -1,13 +1,29 @@
...
+          <Order orderType="pizza" />{" "}
+        </li>
+
+        <li>
+          <Order orderType="salad" />
+        </li>
         <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
+          {" "}
+          <Order orderType="cake" />
```
timhamrouge, line 26: 👍  - Looks good here, 3 separate components for the different Order types! Is there a way we could create all 3 components without having to declare the same component 3 times?

src/SearchResults.jsx
```diff
@@ -0,0 +1,59 @@
...
+            <th scope="col">Check in date</th>
+            <th scope="col">Check Out Date</th>
+            <th scope="col">Number Of Nights</th>
+          </tr>
+        </thead>
+        {props.results.map(customer => {
+          return (
+            <tbody>
+              <tr
+                className={isSelected ? "yellow-background" : null}
```
timhamrouge, line 33: Well done for getting the conditional rending of the class, it selects and deselects as it should. However, it does highlight and un-highlight every row when any of the rows are clicked, where could we apply this functionality so that only one row is selected when it is clicked, instead of all of them?

src/RestaurantButton.jsx
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from "react";
+
+const RestaurantButton = props => {
+  const [orders, countOrder] = useState(0);
+  const incrementOrders = () => {
+    countOrder(orders + 1);
+  };
```
timhamrouge, line 7: 👍  Great stuff with this functionality again. Be careful with the names the exercises are asking for, this code works and if the naming makes sense to you as you're writing it then happy days, but if you're using your own names then make sure they're semantic, i.e.
```
  const incrementOrders = () => {
    countOrder(orders + 1);
  };
```

incrementOrders is a good name (incrementOrder would probably be better because it's only incrementing a single order), but we're not counting anything with the countOrder function, have a think about naming convention for the state variables

src/RestaurantButton.jsx
```diff
@@ -0,0 +1,23 @@
...
+        flexWrap: "wrap"
+      }}
+    >
+      TotalOrders:{orders}
+      <button onClick={incrementOrders}>Increment Orders</button>
+    </div>
+  );
+};
+
+export default RestaurantButton;
```
timhamrouge, line 23: This is the component I was talking about above ^^ You could import this in your Order component and use it there as the functionality is the same as the buttons there. So you have a little bit of repetition and an extra button at the bottom of your app, but the functionality in both places is good and you are using the hooks correctly and effectively so well done 🥇

src/SearchResults.jsx
Taslima26: I tried to work it out but really struggled with it.Thank you

Taslima26:
Thank you Tim for looking at my work and giving me useful suggestions.I will try to improve all of them on next submission .Thank you.

PR #165, opened 2020-07-27
Buchra Atkeh lesson3  Homework

Buchraatkeh1984:


src/CustomerProfile.js
```diff
@@ -0,0 +1,28 @@
...
+  console.log(`https://cyf-react.glitch.me/customers/${customerId}/`);
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${customerId}/`)
+      .then(res => res.json())
+      .then(data => {
+        setCustomerProf(data);
+        console.log(data);
+      });
+  }, [customerId]);
+  if (!customerProf) return "Loading...";
```
dsobkow, line 14: `customerProf` variable is initialised with an empty object so it will always be `truthy` and "Loading" will be never displayed. If you haven't heard about truthy and falsy values in JavaScript - you can read more [here](http://adripofjavascript.com/blog/drips/truthy-and-falsy-values-in-javascript.html). Because an empty object is truthy, you will need to explicitly check that it's empty, for example by checking the length of object keys array:
```
if (!Object.keys(customerProf).length)
```
Also "Loading" message should be an HTML element.

src/OrderChocolatCake.js
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from "react";
+
+import RestaurantButton from "./RestaurantButton";
+
+const OrderChocolateCake = () => {
+  //const types=["Pizza","Salad","Chocolate cake"]
```
dsobkow, line 6: You are not using this component anywhere so the file can be deleted - don't keep unused code🙂

src/OrderSalad.js
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from "react";
+
+import RestaurantButton from "./RestaurantButton";
+
+const OrderSalad = () => {
+  //const types=["Pizza","Salad","Chocolate cake"]
+  const [order, setOrder] = useState(0);
+
```
dsobkow, line 8: This file can also be deleted

src/Restaurant.js
```diff
@@ -1,17 +1,16 @@
...
-    </div>
-  );
-};
+const Restaurant = () => (
+  <div className="orders">
+    <h3>Restaurant Orders</h3>
+    <ul>
+      <Order orderType="Pizzas" />
+      <Order orderType="Salads" />
+      <Order orderType="Chocolate cake" />
```
dsobkow, line 11: Very nice refactoring 👍

src/Search.js
```diff
@@ -1,23 +1,37 @@
...
+  function handleInputChange(event) {
+    setSearchInput(event.target.value);
+    console.log(event.target.value);
+  }
+  function handleSubmit(event) {
+    event.preventDefault();
+    search(searchInput);
+    // setSearchInput(event.target.value);
+    console.log(searchInput);
+    setSearchInput("");
```
dsobkow, line 14: Very good implementation of search input! You even thought about resetting the input value after submitting👏

More console logs and commented out code in this file - please delete it🙂

src/SearchResults.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from "react";
+import HeadingRow from "./HeadingRow";
+import BodyRow from "./BodyRow";
+import CustomerProfile from "./CustomerProfile";
+
+const SearchResults = props => {
+  const [customerId, setCustomerId] = useState(null);
+
```
dsobkow, line 8: This is a very nice and clean component, well done!

src/data/orderTypes.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "id": 1,
```
dsobkow, line 3: Looks like this data is not used anywhere

src/SearchResults.js
```diff
@@ -0,0 +1,88 @@
...
+import HeadingRow from "./HeadingRow";
+import BodyRow from "./BodyRow";
+import CustomerProfile from "./CustomerProfile";
+import moment from "moment";
+
+const SearchResults = ({ results, setBooking }) => {
+  const [customerId, setCustomerId] = useState(null);
+  const [isSorted, setIsSorted] = useState({
+    title: false,
+    firstName: false,
```
dsobkow, line 11: You could even initialise `isSorted` as an empty object and your `if` condition on line 26 would still work. The `isSorted[key]` object property would be undefined and `undefined` is a falsy value.

src/SearchResults.js
```diff
@@ -0,0 +1,88 @@
...
+    let key = event.target.title;
+
+    console.log(isSorted[key]);
+    if (!isSorted[key]) {
+      if (key == "roomId") {
+        sortedBookings = results.sort((a, b) => a[key] - b[key]);
+      } else if (key == "numOfNight") {
+        sortedBookings = results.sort(
+          (a, b) =>
+            moment(a.checkOutDate).diff(a.checkInDate, "days") -
```
dsobkow, line 32: You're calculating the number of nights twice in your app - here and in the `BodyRow` component again so there is some unnecessary code repetition. You could add a `numOfNight` property to each element of the bookings array earlier and then just render it or use for sorting without having to calculate it. Try to think where would be the best place to do that🙂

src/SearchResults.js
```diff
@@ -0,0 +1,88 @@
...
+      const updatedkeystatus = {
+        ...isSorted,
+        [key]: true
+      };
+
+      setIsSorted(updatedkeystatus);
+      console.log(updatedkeystatus);
+      setBooking([...sortedBookings]);
+    } else {
+      if (key == "roomId") {
```
dsobkow, line 46: I think that sorting logic from line 46 to 55 could be extracted to a separate function that takes the `key` as argument and returns a sorted array. `sortBookings` function is quite long and is responsible for many things so keeping the clean code principles in mind - it would be could to break down the responsibilities into smaller functions to improve readability.

PR #173, opened 2020-08-01
Week-3 React

Taslima26:


src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+    email: "",
+    roomId: " ",
+    checkInDate: "",
+    checkOutDate: ""
+  };
+
+  const [customer, setCustomer] = useState(initUser);
+
+  const handleChange = event => {
+    const updateCustomer = {
```
Dedekind561, line 19: This `handleChange` looks spot on well done :)

src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+    if (
+      customer.email &&
+      customer.firstName &&
+      customer.surname &&
+      customer.title &&
+      customer.roomId &&
+      customer.checkInDate &&
+      customer.checkOutDate
+    ) {
+      handleChange(event, props.addCustomer(customer));
```
Dedekind561, line 38: Not sure why you're invoking `handleChange` inside of the `handleSubmit` function. Also `handleChange` is designed to take 1 argument and you're passing 2 arguments.

src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+      [event.target.name]: event.target.value
+    };
+    setCustomer(updateCustomer);
+  };
+
+  const handleSubmit = event => {
+    event.preventDefault();
+    console.log(customer.email);
+    if (
+      customer.email &&
```
Dedekind561, line 30: Excellent piece of validation to check if the user has filled in all the fields of the form 😄

src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+      customer.roomId &&
+      customer.checkInDate &&
+      customer.checkOutDate
+    ) {
+      handleChange(event, props.addCustomer(customer));
+    }
+  };
+
+  return (
+    <form class="form-group">
```
Dedekind561, line 43: If you create a form element then you need to attach the `handleSubmit` event handler function to the form element

Dedekind561: You've attached it to a div  below

src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+          value={customer.checkInDate}
+          name="checkInDate"
+          onChange={handleChange}
+        />
+      </div>
+      <div>
+        <label>CheckOut</label>
+        <input
+          className="u-full-width"
+          type="date"
```
Dedekind561, line 108: Nice use of `type` on the check in/out fields

src/components/AddCustomerForm.jsx
```diff
@@ -0,0 +1,123 @@
+import React, { useState } from "react";
+import "./AddCustomerForm.css";
+
+const AddCustomerForm = props => {
```
Dedekind561, line 4: Could be a good idea to destructure the `addCustomer` function here too

src/components/Bookings.js
```diff
@@ -0,0 +1,69 @@
+import React, { useState, useEffect } from "react";
+import Search from "./Search.js";
+import SearchResults from "./SearchResults.jsx";
+import AddCustomerForm from "./AddCustomerForm";
+import "./Bookings.css";
+
+const Bookings = () => {
+  const [Booking, setBookings] = useState([]);
```
Dedekind561, line 8:
```suggestion
  const [bookings, setBookings] = useState([]);
```

src/components/Bookings.js
```diff
@@ -0,0 +1,69 @@
...
+  const [error, setError] = useState(null);
+
+  console.log(loading);
+  console.log(Booking);
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/delayed`)
+      .then(res => res.json())
+      .then(data => {
+        if (data.error) {
+          throw data;
```
Dedekind561, line 19: Here you're throwing an error if the api request is unsuccessful and this is problematic for 2 reasons:
- The `throw` means the app will crash and stop working
- You won't hit the other bit of error handling you did in your `catch` block

src/components/Bookings.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        setBookings(data);
+        setLoading(false);
+      })
+      .catch(error => setError(error));
```
Dedekind561, line 25: Good error handling here - this is the place that you want to handle the errors

src/components/Bookings.js
```diff
@@ -0,0 +1,69 @@
...
+          <div>
+            <p>Data is loading...</p>
+          </div>
+        ) : (
+          <div>
+            <SearchResults results={Booking} />
+          </div>
+        )}
+      </div>
+      <div className="container">
```
Dedekind561, line 60: At the moment you've got logic for the loading to be displayed - which is great. Need to ensure you've got the same when an error occurs on the page too

src/components/Bookings.js
```diff
@@ -0,0 +1,69 @@
...
+      })
+      .catch(error => setError(error));
+    console.log(error);
+  }, []);
+
+  console.log(loading);
+  console.log(Booking);
+  const search = searchVal => {
+    console.info("TO DO!", searchVal);
+    const filteredBooking = Booking.filter(
```
Dedekind561, line 33: Good variable name here for `filteredBooking`

src/components/CustomerProfile.jsx
```diff
@@ -0,0 +1,22 @@
...
+  const [customerData, setCustomerData] = useState({});
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then(res => res.json())
+      .then(data => setCustomerData(data));
+  }, [id]);
+  console.log(customerData);
+  return (
+    <div>
+      <p>Customer {id} profile </p>
```
Dedekind561, line 13: This looks good on the page should be an `ul` with nested `li` elements

src/components/SearchResults.jsx
```diff
@@ -0,0 +1,104 @@
...
+import Bookings from "./Bookings";
+import CustomerProfile from "./CustomerProfile";
+import TableRow from "./TableRow";
+function SearchResults({ results }) {
+  const [customerId, setCustomerId] = useState(null);
+  const sortedResults = [...results];
+  const [sortedField, setSortedField] = useState(null);
+  if (sortedField !== null) {
+    sortedResults.sort((a, b) => {
+      if (a[sortedField] < b[sortedField]) {
```
Dedekind561, line 12: `sort` will default to sorting by ASCII code so you may want to ensure you lower case the items in the array before comparing them.

src/components/SearchResults.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+import Bookings from "./Bookings";
+import CustomerProfile from "./CustomerProfile";
+import TableRow from "./TableRow";
+function SearchResults({ results }) {
+  const [customerId, setCustomerId] = useState(null);
+  const sortedResults = [...results];
+  const [sortedField, setSortedField] = useState(null);
+  if (sortedField !== null) {
+    sortedResults.sort((a, b) => {
```
Dedekind561, line 11:
```suggestion
    sortedResults.sort((result1, result2) => {
```

src/components/SearchResults.jsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+  }
+
+  return (
+    <div>
+      <table className="table">
+        <thead className="thead-dark">
+          <tr>
+            <th scope="col">
+              <button type="button" onClick={() => setSortedField("id")}>
```
Dedekind561, line 28: Some of the columns are mixed up here - I think you mixed around the `title` and `firstName` in the table

src/components/SearchResults.jsx
```diff
@@ -0,0 +1,104 @@
+import React, { useState } from "react";
+
+import Bookings from "./Bookings";
+import CustomerProfile from "./CustomerProfile";
+import TableRow from "./TableRow";
+function SearchResults({ results }) {
+  const [customerId, setCustomerId] = useState(null);
+  const sortedResults = [...results];
```
Dedekind561, line 8: `sortedResults` is being mutated by the `sort` method so beware of calling a mutating method on this array

src/components/AddCustomerForm.jsx
Taslima26: Yes probably no need to pass an event here.Thank you.

src/components/AddCustomerForm.jsx
Taslima26: Aah that was by mistake thats why I was getting some strange behaviour.

src/components/SearchResults.jsx
Taslima26: Sort was difficult to implement I know its still not working as it should be.

Taslima26:
Thank you for looking at my work .I will go through changes you have suggested.
Thank you

PR #174, opened 2020-08-03
react week1 homework

alastair87:


src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+      </thead>
+      <tbody>
+        {props.results.map(result => {
+          return (
+            <tr key={(Math.random() / 10000000000).toString()}>
+              {Object.keys(result).map(function(key) {
+                return (
+                  <td key={(Math.random() * 10000000000).toString()}>
+                    {result[key]}
+                  </td>
```
alastair87, line 27: I am not sure what you are trying to do by using a random number generator to give the rows and data cells keys, however you don't need to do anything complicated as the bookings each have a unique id field. You can use this value for your key. You also shouldn't need to give each cell a unique identifier, as it is already grouped with the rest of the data for that booking.

src/SearchResults.js
```diff
@@ -0,0 +1,46 @@
...
+  );
+};
+
+function NumberNights(result) {
+  let checkIn = moment(result.checkInDate);
+  let checkOut = moment(result.checkOutDate);
+
+  return checkOut.diff(checkIn, "days");
+}
+
```
alastair87, line 45: Good. One thing you can try to start doing is take note of when when you're creating variables that will never be altered or replaced (within that scope). In this case you set checkIn and checkOut once each and never need to change them again. This means you *could* use `const` rather than `let`. This doesn't change what the code does, but doing so can have the advantage of making it easier to spot if you are changing or replacing a value unintentionally later in your code as you'll get an error.

PR #209, opened 2020-10-07
Mohammed React Week 1 Solution

mhallaq:


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "moments": "0.0.2",
```
peedeerich, line 6: this may be a typo :) I think you mean moment.

This is a real npm module! And it has `moment` as a dependency - this is why your code is still working correctly but in production it is important to make sure the correct modules are loaded.

src/Bookings.js
```diff
@@ -12,6 +15,7 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
+        <SearchResults searchReslt={FakeBookings} />
         {/* <SearchResults results={FakeBookings} /> */}
```
peedeerich, line 19: it is good practice to delete comments like this once you no longer need them

src/TouristInfoCards.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+const TouristInfoCards = () => (
```
peedeerich, line 3: each of these cards have the same format so it would be good if either a) they were a component or b) use a `map` function to take the data (city name, image, link url etc) and return the card

peedeerich (commented):
nice work :) See comments below

PR #250, opened 2021-04-18
London Class 7 - Ionut Tripa - React - Week 1 - CYF Hotel

iTrip23:
Your Name: Ionut Tripa
    Your City: London
    Your Slack Name: Johnny


src/SearchButton.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
```
louisechow, line 1: Excellent job extracting the search button. Good choice of name for you component and good to see you use PascalCase for your component and file names.

src/Heading.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const Heading = () => (
+  <header className="App-header">
```
louisechow, line 4: Good job extracting the heading!

One small suggestion; the most common convention for naming css classes is to use all lower case and have `-` between words. Most of your classes in `App.css` follow this convention except for the two classes used here. I would change them to `app-header` and `app-logo` so your code is consistent.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,78 @@
+import React from "react";
+
```
louisechow, line 2: This looks great on your page.

Notice though that you have very similar code for each of the three cards since they all have the same structure (all the HTML elements are the same, it's just the content that changes). When you see code repeated, that's usually a good sign that you can simplify it by creating a separate component.

Why don't you try creating a separate `<Card>` React component and store the city data in an array that you can then iterator over to create a new card component for each city. Similar to the [example that you did in class](https://codesandbox.io/s/interpolation-with-map-in-jsx-7mw0mw3qx0?file=/src/MentorsList.js).

src/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
```
louisechow, line 2: This looks good! Great to see you added the `key` to each element as you iterated through the array.

A very small suggestion. Since this is a footer component on the page, it would make semantic sense for you to wrap the `<ul>` element inside a `<footer>` element and move the class onto this new `<footer>` element. It won't change any visually on the screen, but improve the accessibility of your page (screen readers for people who are blind rely on semantics in pages to describe the page better for them).

src/SearchResults.js
```diff
@@ -0,0 +1,36 @@
+import moment from "moment";
```
louisechow, line 1: You forgot to import `React` in this component, so it failed to load on the page.

src/SearchResults.js
```diff
@@ -0,0 +1,36 @@
...
+                <th scope="col">Surname</th>
+                <th scope="col">email</th>
+                <th scope="col">Room ID</th>
+                <th scope="col">Check In Date</th>
+                <th scope="col">Check Out Date</th>
+                <th scope='col'>Number of nights stay</th>
+            </tr>
+        </thead>
+        <tbody>
+            {props.results.map(guest => (
```
louisechow, line 19: Don't forget that when you iterate over an array, you'll need to make sure each element has a `key`. Right now there are errors throwing in the browser console because of this.
https://syllabus.codeyourfuture.io/react/week-1/lesson#keys

src/SearchResults.js
```diff
@@ -0,0 +1,36 @@
...
+                <tr>
+                    <td>{guest.id}</td>
+                    <td>{guest.title}</td>
+                    <td>{guest.firstName}</td>
+                    <td>{guest.Surname}</td>
+                    <td>{guest.email}</td>
+                    <td>{guest.roomId}</td>
+                    <td>{guest.checkInDate}</td>
+                    <td>{guest.checkOutDate}</td>
+                    <td>{moment(guest.checkOutDate).diff(moment(guest.checkInDate))}</td>
```
louisechow, line 29: The values in the `Number of nights stay` columns are currently showing in milliseconds instead of day. You will need to pass in a second argument to the `diff` function to tell it to return the number in days.
See the documentation at https://momentjscom.readthedocs.io/en/latest/moment/04-displaying/07-difference/

src/App.css
```diff
@@ -4,18 +4,20 @@
```
louisechow, line 4: One thing that would be good for you to start doing is to create a separate css files for each of your React components and putting the related classes in that.

E.g. for the `Heading.js` component, create a `Heading.css` file and put the `.app-logo` and `.app-header` classes into that and import that file into `Heading.js`
```
import "./Heading.css";
```

Why do this? That way you don't end up with one huge css file that contains everything. It's also easier to find which classes apply to which components.

louisechow:
Great job with the homework this week Johnny! 

While you are developing, don't forget to check the browser console to see if your code is throwing any errors. You would be able to catch some of the issues by checking those messages. Let me know if you have any questions about the feedback I left.

PR #259, opened 2021-04-22
London Class 7 - Daniel Ghanbari - React - Week 1 - CYF Hotel

DanGhanbari:


src/SearchButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
```
louisechow, line 1: Excellent job extracting the search button. Good choice of name for your component and good to see you use PascalCase for your component and file names.

src/Heading.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+const Heading = () => {
```
louisechow, line 3: Good job extracting the heading!

Some small suggestions; 
- The most common convention for naming css classes is to use all lower case and have - between words. I would change the class name you use here to `app-header` to follow the convention.
- On line 10-11 you have styling logic set on the image directly. You should try to move all styling logic into CSS and in your JSX/HTML just add classes to indicate what styles should apply. So in this case, I would create a new CSS class and then add a the class on the `img` tag.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+
+const TouristInfoCards = () => {
+  return (
```
louisechow, line 4: This looks great on your page.

Notice though that you have very similar code for each of the three cards since they all have the same structure (all the HTML elements are the same, it's just the content that changes). When you see code repeated, that's usually a good sign that you can simplify it by creating a separate component.

Why don't you try creating a separate <Card> React component and store the city data in an array that you can then iterator over to create a new card component for each city. Similar to [the example that you did in class](https://codesandbox.io/s/interpolation-with-map-in-jsx-7mw0mw3qx0?file=/src/MentorsList.js).

src/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
```
louisechow, line 3: This looks great! Great to see you added the key to each element as you iterated through the array.

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
+import moment from "moment";
+import React from "react";
```
louisechow, line 2: Great job with this component! 👏

src/App.css
```diff
@@ -1,15 +1,17 @@
 .App {
```
louisechow, line 1: One thing that would be good for you to start doing is to create a separate css files for each of your React components and putting the related classes in that.

E.g. for the Heading.js component, create a Heading.css file and put the `.app-header` class into that and import that file into Heading.js
```
import "./Heading.css";
```
Why do this? That way you don't end up with one huge css file that contains everything. It's also easier to find which classes apply to which components.

louisechow:
Good job with the homework this week Daniel! Good to see your code was well formatted and there were no errors in the console from your code. Let me know if you have any questions about the feedback I left.

DanGhanbari:
Hi Louise,
Thank you for your feedback. I'll try to fix those mistakes.

src/Bookings.js
```diff
@@ -1,18 +1,32 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
+import SearchResults from "./SearchResults.js";
 // import FakeBookings from "./data/fakeBookings.json";
 
+const [bookings, setBookings] = useState([]);
```
louisechow, line 6: Right now when I run your code I get an error saying:
```
Hooks can only be called inside of the body of a function component.
```
Line 6 needs to be inside the `Bookings` component.

src/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+      <ul className="list-group mx-2">
+        <Order orderType={"Pizzas"} />
+        <Order orderType={"Salads"} />
+        <Order orderType={"Chocolate cake"} />
       </ul>
```
louisechow, line 12: For lines 9 - 11, you don't need to use the curly braces since you are just passing in a hard-coded string.
You only need to use `{ }` if you are passing it some javascript.

src/Restaurant.js
```diff
@@ -1,14 +1,14 @@
...
 const Restaurant = () => {
-  const pizzas = 0;
   return (
-    <div>
+    <div className="restaurant-orders my-4">
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
```
louisechow, line 11: Good job splitting out the `Order` component!

src/SearchResults.js
```diff
@@ -0,0 +1,51 @@
+import moment from "moment";
+import React, { useState } from "react";
+
+const SearchResults = props => {
+  const [color] = useState("");
```
louisechow, line 5: On line 5, you have not declared the `set` method for the state and you never call it. Because of this you are not using the power of state to trigger the re-render of your component for you automatically.

I think you can make a change to the way you have implemented the row highlighting to use state better.

My suggestion, create a new component `SearchResultRow` that contains lines 31- 43, and replace those lines with your new component
```
<tbody className="text-center">
          {props.results.map((person, index) => (
            <SearchResultRow key={index} person={person} />
          ))}
</tbody>
```

Move both your `state` and `onClickHandler` into this new component. Now, for each instance of your SearchResultRow (aka for each row), it will have it's own state keeping track of whether it is selected or not.

Now, whenever `onClick` is called, all your function has to do is `set` the status to indicate if the row is highlighted or now. When the `set` method is called, it will automatically trigger a re-render.

You can then use the state to change the `className` using a ternary operator. Something like
```
<tr onClick={highlightRow} className={isSelected ? "gray" : ""}>
```

src/Bookings.js
```diff
@@ -1,18 +1,32 @@
...
+
 const Bookings = () => {
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    let filterBookings = bookings.filter(booking => {
```
louisechow, line 11: Good job filtering the bookings here!

src/Search.js
```diff
@@ -1,23 +1,35 @@
-import React from "react";
+import React, { useState } from "react";
+import SearchButton from "./SearchButton";
 
-const Search = () => {
+const Search = props => {
```
louisechow, line 4: Your state for the changing input and submission both look good!

PR #260, opened 2021-04-22
London 7 - Hozan Ali - React - Week 1 - CYF Hotel

Hozan94:


src/App.js
```diff
@@ -1,13 +1,35 @@
...
+  "hello@fakehotel.com",
+  "0123 456789"
+];
+
 const App = () => {
+  const [newBooking, setNewBooking] = useState("");
+
+  function getNewBooking(booking) {
+    setNewBooking(booking);
+  }
```
40thieves, line 23: I'm not sure this is exactly what we want. At the moment, I think we can only add 1 booking. If you try to add a second booking, nothing happens.

What could we do to fix that? Hint: what we actually want to do is add the new booking to our existing list of bookings. So how could we do that?

src/TableRow.js
```diff
@@ -0,0 +1,44 @@
...
+    setHighlight(!highlight);
+  }
+
+  function handleClick() {
+    props.searchId(props.customer.id);
+  }
+
+  return (
+    <tr
+      key={props.customerIndex}
```
40thieves, line 17: I think this is actually unnecessary 🙂 Try removing it and look to see if you get a warning in the dev tools console.

src/SearchResults.js
```diff
@@ -0,0 +1,90 @@
...
+                />
+              );
+            })}
+          </tbody>
+          <tfoot>
+            {props.addedBooking ? (
+              <NewBookingsTableRow
+                customer={props.addedBooking}
+                searchId={search}
+              />
```
40thieves, line 80: Is there any difference between this component and `TableRow`? If not, what could we do to tidy up the code a bit?

40thieves (commented):
Really well done @Hozan94!! ⚡ 🏎️ 🎉

src/App.js
Hozan94: @40thieves Many thanks for the feedback, I updated it and it works now the way it should be.

src/SearchResults.js
```diff
@@ -0,0 +1,90 @@
...
+
+    function search(id) {
+      //added an if else to collapse the list (customer info)
+      if (customerId === id) {
+        setCustomerId(false);
+      } else if (customerId) {
+        setCustomerId(id);
+      } else {
+        setCustomerId(id);
+      }
```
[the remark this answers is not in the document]
Hozan94: @40thieves You are right it is not necessary to have both, so I deleted the `else if` statement.

src/LoadingRow.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+
+function TableRow(props) {
```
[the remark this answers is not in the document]
Hozan94: @40thieves you are right I have the file and the functions with different names, but if I was exporting the function `TableRow` how come I did not get any error when In `SearchResults.js` I have imported `LoadingRow` component (which does not exist) from the `LoadingRow` file which contains only `TableRow` component?

Also regarding `LoadingRow`, thanks for this note as well, it is a whole table not just a row.

src/Search.js
```diff
@@ -1,23 +1,38 @@
-import React from "react";
+import React, { useState } from "react";
+import SearchButton from "./SearchButton";
+
+const Search = props => {
+  const [searchInput, setSearchInput] = useState("");
+
+  function handleSearchInput(event) {
+    //Can also be written as an arrow func inside `onChange`
```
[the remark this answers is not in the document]
Hozan94: @40thieves is it becuase we can use the same callback function when there is another event handler that wants `event.target.value` ?

PR #262, opened 2021-04-23
London 7 - Benyam Worku- React - Week 1 - CYF Hotel

BenyamWorku:
I haven't figured out how to use 
1. A helper function that is defined outside the components so I put the function inside the component
2. the moment library so I used new Date and took the difference


src/SearchResults.js
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable indent*/
+import React, { useState } from "react";
+
+import bookingData from "./data/fakeBookings.json";
+function LengthOfStay(endDate, startDate) {
+  const diff = new Date(endDate) - new Date(startDate);
+  return Math.floor(diff / 86400000);
+}
```
Jpsi, line 8: Good way of finding the difference between two dates. 👍

Tips for more readable code:
- use `differenceInMilliseconds` instead of `diff`
- define `const millisecondsInOneDay = 24*3600*1000;` and return `Math.floor(differenceInMilliseconds / millisecondsInOneDay`

BenyamWorku: > Good way of finding the difference between two dates. 👍
> 
> Tips for more readable code:
> 
> * use `differenceInMilliseconds` instead of `diff`
> * define `const millisecondsInOneDay = 24*3600*1000;` and return `Math.floor(differenceInMilliseconds / millisecondsInOneDay`

Thanks Philip. I've changed it now.

BenyamWorku:
> I haven't figured out how to use
> A helper function that is defined outside the components so I put the function inside the component

Figured it out now.

src/Bookings.js
```diff
@@ -1,19 +1,39 @@
...
+      .then(data => {
+        data = data;
+        setBookings(data => {
+          return data.filter(elem =>
+            [elem.firstName.toLowerCase(), elem.surname.toLowerCase()].includes(
+              searchVal.toLowerCase()
+            )
+          );
+        });
+      });
```
beckyf0keefe, line 28: This is a really interesting way of approaching this, combining the search and the fetch, it works well here! Just be cautious to ensure that it doesn't become too complicated, readability of your code should be always a priority

src/Order.js
```diff
@@ -0,0 +1,19 @@
...
+    const newQuantity = orders + 1;
+    setOrder(newQuantity);
+  }
+  return (
+    <li className="order-list">
+      {props.orderType}:{orders}
+      <RestaurantButton update={addOne} />
+    </li>
+  );
+}
```
beckyf0keefe, line 17: This is a great use of props to create a reusable component, to make this even better, have you tried adding 1 to the order number within the setOrder() function? This would remove the need for the newQuantity constant completely!

src/RestaurantButton.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable indent*/
+import React from "react";
+// import Restaurant from "./Restaurant";
+
+function RestaurantButton(props) {
+  return (
+    <button
+      className="add-btn"
+      onClick={props.update}
+      className="btn btn-primary"
```
beckyf0keefe, line 10: Good to see you passing functions as props, this is a step up from passing simple information, to make this even better try naming your props in a way that you can tell what is does straight away, also try to just add all your classNames in one line to keep you code as readable as possible

PR #267, opened 2021-04-25
London-Class 7-Murat Demirtas-React week 1-CFY Hotel

muratdemirtas2001:


src/BookingForm.js
```diff
@@ -0,0 +1,127 @@
...
+        <h3>Add a new book</h3>
+      </div>
+      <div className="form-element">
+        <input
+          type="text"
+          required
+          name="title"
+          placeholder="Title"
+          value={title}
+          onChange={handleFormChange}
```
Buroni, line 57: I think your solution is good, but another way to do it would be to use the state functions directly in `onChange`. For example

```js
<input name="title" onChange={(event) => setTitle(event.target.value)} />
```

src/Bookings.js
```diff
@@ -1,21 +1,155 @@
...
+          return resp.json();
+        }
+      })
+      .then(data => {
+        setBookings(data);
+        setLoading(false);
+      })
+      .catch(error => {
+        console.log(error);
+        setError(error.message);
```
Buroni, line 38: Nice use of promise chaining 👍

src/Bookings.js
```diff
@@ -1,21 +1,155 @@
...
-  return (
-    <div className="App-content">
-      <div className="container">
-        <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+  function columnSorter(e) {
+    let id = parseInt(e.target.id);
+    if (id === 0) {
+      if (columnClicked[0] === false) {
+        setColumnClicked([true, false, false, false]);
```
Buroni, line 55: This solution works, but the logic is confusing. I don't think it's necessary to store 4 booleans, since it's only possible for one column to be sorted at one time. Personally I would store the `columnClicked` as a number which stores the ID of the clicked field. So `setColumnClicked(id);`

src/Bookings.js
```diff
@@ -1,21 +1,155 @@
...
+      });
+      newBookings = [...newBookings];
+      setBookings(newBookings);
+    } else if (id === 3) {
+      setColumnClicked([false, false, false, true]);
+      let newBookings = bookings.sort((book1, book2) => {
+        return book1.email.localeCompare(book2.email);
+      });
+      newBookings = [...newBookings];
+      setBookings(newBookings);
```
Buroni, line 94: You're writing `setBookings` inside every `if` statement; why not just write it once at the end?

Also, this is a bit harder, but can you think of a way to only write `bookings.sort` once instead of inside every `if` statement?

src/SearchResults.js
```diff
@@ -0,0 +1,41 @@
...
+        <td>{results.firstName}</td>
+        <td>{results.surname}</td>
+        <td>{results.email}</td>
+        <td>{results.roomId}</td>
+        <td>{results.checkInDate}</td>
+        <td>{results.checkOutDate}</td>
+        <td>
+          {moment(results.checkOutDate).diff(
+            moment(results.checkInDate),
+            "days"
```
Buroni, line 28: Nice use of moment-js 👍

Buroni (commented):
Hey Murat, good job! The app works well when tested, and the tasks have been completed. You've also used `flex` well to align all of your content, and your work shows a good understanding of state management!

I think it may be worth it to re-visit your `columnSorter` function, as there are some ways the logic could be improved and made clearer. I'm happy to answer any questions about that or go through the code together on call!

src/TouristInfoCards.js
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+
+function TouristInfoCards() {
+  return (
+    <div className="cards-wrapper">
+      <div className="card info-card">
+        <img
+          src="https://images.unsplash.com/photo-1546867042-cdba19743124?ixid=MnwxMjA3fDB8MHxzZWFyY2h8MjkxfHxnbGFzZ293fGVufDB8fDB8fA%3D%3D&ixlib=rb-1.2.1&auto=format&fit=crop&w=500&q=60"
+          className="card-img-top"
+          alt="Glasgow"
```
Buroni, line 10: If you want to make this code shorter and easier to maintain, you could use a `map` instead of writing the same HTML structure three times.

```js
const cityInfo = [
    {name: "Glasgow", text: "Glasgow is the largest city in Scotland and ...", moreInfoLink: "http://www.peoplemakeglasgow.com", imgSrc: "https://images.unsplash.com/photo-1546867042"},
    {name: "Manchester", text: "Manchester is one of the ...", moreInfoLink: "http://www.visitmanchester.com", imgSrc: "https://images.unsplash.com/photo-1546867042"},
    // ...
];
```

```js
return (
    <div className="cards-wrapper">
        { cityInfo.map(city => /* City HTML here */)  }
    </div>
)
```

PR #268, opened 2021-04-25
London - Class 7 - React - CYF Hotel

iTrip23:
Your Name: Ionut Tripa
    Your City: London
    Your Slack Name: Johnny


src/components/Restaurant.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import Order from "./Order.js";
+
+const Restaurant = () => {
+  return (
```
louisechow, line 5: Great job here breaking out the `Order` component

src/components/Order.js
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton.js";
+
```
louisechow, line 3: Good formatting and style! Your code looks nice and clean!

src/components/GuestRow.js
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+const GuestRow = ({ guest, setGuestID }) => {
+  const [highlight, setHighlight] = useState(false);
+
+  const Toggle = () => setHighlight(!highlight);
```
louisechow, line 7: Very small nitpick, but i'd choose a different name instead of `Toggle`. 

Firstly, since it's a function, you should name it starting with a lower case. When you use an upper case it looks like it's a component (and it's even highlighted as such).

`toggle` is a great action verb to use, but I think you could make it more descriptive and name it `toggleSelectedRow`. But that's probably just my personal preference.

src/components/Search.js
```diff
@@ -1,23 +1,35 @@
...
-const Search = () => {
   return (
     <div className="search">
       <div className="page-header">
         <h4 className="text-left">Search Bookings</h4>
       </div>
       <div className="row search-wrapper">
         <div className="col">
-          <form className="form-group search-box">
+          <form className="form-group search-box" onClick={handleSubmit}>
```
louisechow, line 21: You should use the `onSubmit` for `form` elements instead of `onClick`. 

Although it will behave the same in this case, in other cases where you have a long form with many inputs,  `onSubmit` will be the event that is used to trigger validation of your input fields.

src/components/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
...
+      </p>
+      <p>
+        Guest Phone Number:<span>{profileID.phoneNumber}</span>
+      </p>
+      <p>
+        Guest Email:<span>{profileID.email}</span>
+      </p>
+    </section>
+  ) : (
+    <h3>Loading</h3>
```
louisechow, line 26: This `Loading` message shows on the page before the user clicks on a profile, which is confusing.

I think you want to move the display of the `Loading` text a level up to the `SearchResults` component instead. For this component, instead of using the ternary operator, i'd use the logical and to render only if there are profile details.

E.g.
```
return profileID !== null && (
    <section className="customer-profile">
     .......
    </section>
  )
```

src/components/SearchResults.js
```diff
@@ -0,0 +1,56 @@
...
+            <th scope="col">Customer Info</th>
+          </tr>
+        </thead>
+        <tbody>
+          {filter.map((guest, index) => (
+            <GuestRow guest={guest} key={index} setGuestID={setGuestID} />
+          ))}
+        </tbody>
+      </table>
+      <CustomerProfile profileID={profileID} guestID={guestID} />
```
louisechow, line 51: This would be the place where you can choose to either display `Loading...` text or to render the `CustomerProfile` component.

Why don't you add a new `isLoading` state in this component and set the state appropriately in your `useEffects` function. Then you can use the state to choose whether to display the loading text or the `CustomerProfile`.

src/components/Bookings.js
```diff
@@ -0,0 +1,25 @@
+import React, { useState, useEffect } from "react";
+import Search from "./Search.js";
+import SearchResults from "./SearchResults.js";
+
+const Bookings = () => {
+  const [bookings, setBookings] = useState([]);
+  const [submit, setSubmit] = useState("");
+
+  useEffect(() => {
```
louisechow, line 9: Similar to my comments in the `SearchResults` component. You can use state in this component to render appropriate `loading` and `error` messages on your website depending on the state of your API call.

louisechow:
Great job moving your components into the `components` folder. Good structuring of your files in your codebase! 👏

src/components/SearchResults.js
iTrip23: Thank you for the insight, finally managed to implement it

PR #270, opened 2021-04-27
Finished first week

ASLANDENIZ:


src/Bookings.js
```diff
@@ -1,18 +1,42 @@
...
+        .then(res => res.json())
+        .then(data => {
+          setBookings(data);
+          setLoadingData(true);
+        });
+    } catch (error) {
+      setError(true);
+      console.log(error);
+    }
+  }, []);
```
beckyf0keefe, line 22: This useEffect is looking good, just make sure to check the URL that you are fetching from, if the data type that this URL returns is not the same as what you are expecting, it could cause some issues.

src/App.css
```diff
@@ -2,6 +2,10 @@
   text-align: left;
 }
 
+.marked {
+  background-color: red;
+}
```
beckyf0keefe, line 7: Good use of CSS to add a highlight to a row, to make this even better try to make your class names something that anyone reading them could work out what it does, i.e. "highlighted-row" or similar

src/Order.js
```diff
@@ -0,0 +1,18 @@
...
+const Order = ({ orderType }) => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
+    setOrders(orders + 1);
+  };
+  return (
+    <div>
+      <li>
+        {orderType}:{orders} <RestaurantButton orderOne={orderOne} />
+      </li>
```
beckyf0keefe, line 13: Great use of props to create a reusable component!

PR #282, opened 2021-10-04
NW4-Anza-Azam-React-week one (1-7 tasks)

Anza-Azam:


src/components/Restaurant.js
```diff
@@ -1,14 +1,15 @@
...
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order orderType={"Pizzas"} />
+        <Order orderType={"Salads"} />
+        <Order orderType={"ChocolateCake"} />
```
Nomes27, line 12: Small typo here

Nomes27 (commented):
Great work Anza, you show a really good understanding of passing props to components and using components to break up the different sections of your app. There are some small things you could do to help with the presentation of your app, for example, the alignment of your ‘Restaurant Orders’, and some small margins could help with spacing out your restaurant list items. The functionality is great so far, well done!

src/components/Footer.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
-const Footer = prop => {
-  const hotelInfo = prop.hotelContact;
+const Footer = ({ hotelContact }) => {
+  const hotelInfo = hotelContact;
   return (
     <ul>
       {hotelInfo.map((info, index) => (
```
Nomes27: nice destructuring! you could make this more concise by using your hotelContact instead of hotelInfo in your map function (and then can get rid of your hotelInfo variable)

src/components/NewBooking.js
```diff
@@ -65,9 +65,12 @@ function NewBooking(props) {
 
         <input
           type="number"
+          min="6"
+          max="1000"
```
Nomes27, line 69: this works well :)

src/components/Order.js
```diff
@@ -9,7 +9,7 @@ const Order = props => {
     setOrders(orders + 1);
   };
   const subtractOne = () => {
-    setOrders(orders - 1);
+    setOrders(orders === 0 ? orders : orders - 1);
```
Nomes27, line 12: nice, the order number now never goes below 0, well done! Could you think of a way to approach this so that it only sets state if orders is > 0? At the moment state is getting updated even if orders ===0

src/components/SearchResults.js
```diff
@@ -18,12 +18,7 @@ const SearchResults = prop => {
   };
 
   const handleClick = field => {
-    if (
-      field === "firstName" ||
-      field === "title" ||
-      field === "surname" ||
-      field === "email"
-    ) {
+    if (field === "firstName" || field === "title" || field === "surname") {
```
Nomes27, line 21: You could leave in your email field, it's just your map function below on line 22 which converts to lowercase isn't being used anywhere (you aren't currently saving it to a variable to use later on.) In your sort function, you are sorting values which are a mixture of lowercase & uppercase, this is fine for most columns, but your email address column has a mixture of upper & lowercase values so it is not sorting correctly. Send me a message on slack if you need me to explain more :)

PR #285, opened 2021-10-06
hotel-react Maziar Majd

Maziarmajd:


KarenPudner (approved):
Good work - you have met all the other criteria and your deployed app looks correct, apart from a small layout issue with the footer.

KarenPudner (approved):
Good work - I like the design of you deployed app, all the functionality works and you have followed all the instructions.

src/App.css
```diff
@@ -46,12 +50,59 @@ tr {
...
+  margin-top: -200px;
   padding-top: 20px;
-  text-align: center;
+}
+.card:hover {
+  opacity: 2;
+  transform: scale(1.05);
+  box-shadow: 2px 2px 10px 14px rgba(255, 255, 255, 0.2);
+  background-color: rgb(14, 165, 21);
+}
```
Nomes27, line 82: Nice use of on hover properties here! With opacity, the docs state the value can be **A <number> in the range 0.0 to 1.0, inclusive, or a <percentage> in the range 0% to 100%** so will need to have a look at your value for this :)

src/Footer.js
```diff
@@ -0,0 +1,19 @@
...
+  "0123 456789"
+];
+
+const Footer = () => {
+  return (
+    <ul className="footer">
+      {footerInfo.map(info => (
+        <li>{info}</li>
+      ))}
+    </ul>
```
Nomes27, line 15: Remember to use a key whenever you map in React, noticed you have used them elsewhere, which is good!

src/Order.js
```diff
@@ -0,0 +1,19 @@
...
+  //const pizzas = 0;
+
+  function orderOne() {
+    setOrders(orders + 1);
+  }
+  return (
+    <li>
+      <RestaurantButton clickFunc={orderOne} />
+      {props.orderType}: {orders}
+    </li>
```
Nomes27, line 15: Nice passing of the function here. Would be a good idea to change the name of 'clickFunc' to something which is more explanatory (for example orderOne = {orderOne} ) Someone else reading your code may wonder what clickFunc relates to when looking at the child component.

src/SearchResultRow.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from "react";
+import moment from "moment";
+import "./App.css";
+
+function SearchResultRow({ booking }) {
```
Nomes27, line 5: really nice destructuring here!

PR #297, opened 2021-10-13
Scotland Class5 - Megumi Yanaka

ghost:
I completed Lesson1 exercises. 
Gradually I used to a way to create components. 
Now understand how to display data with array.map() method, with props. 
Not understand well about index as value of key.
Still don't have confidence to use Bootstrap and CSS.


src/SearchResults.js
```diff
@@ -13,19 +14,23 @@ function SearchResults(props) {
           <th scope="col">Room id</th>
           <th scope="col">Check in date</th>
           <th scope="col">Check out date</th>
+          <th scope="col">Number of nights</th>
         </tr>
       </thead>
       <tbody>
         {props.FakeInfo.map((info, i) => (
```
filkovsp: it's not always fake :) one day it must become real live results data :)

src/Bookings.js
```diff
@@ -1,18 +1,19 @@
...
 import Search from "./Search.js";
 import SearchResults from "./SearchResults.js";
-import FakeInfo from "./data/fakeBookings.json";
+import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
   const search = searchVal => {
     console.info("TO DO!", searchVal);
   };
+  const [bookings, setBookings] = useState(FakeBookings);
```
filkovsp, line 10: same can be done also this way:
```js
const bookings = useState(FakeBookings);
```
difference is that instead of destructuring whatever pair of "state" and "setter" into an array of variables you can just simply catch only first element from useState() - the state variable.
b-coz if you catch "setter" function but don't use it - react gives that annoying warning in console that setBookings has been declared but never used.

PR #344, opened 2022-03-20
London Class 8 - Berkeli Halmyradov - CYF Hotel React

berkeli:
@gixo This weeks homework

Thanks


src/Header.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export default function Header({ appModes, setAppModehandler }) {
```
gixo, line 3: Just a small nitpick the instructions call for a component named 'Heading', while you have called it 'Header'. I'd make sure you follow the instructions as close as possible when they are so detailed, as this is often an important evaluation criterion for interviews.

gixo (approved):
Well done Berkeli, the code looks very good overall.

I would suggest next to start looking into testing your code more, now that you have added all the components, state transitions, effects, etc. to the page.

How would you use testing libraries like the React Testing Library, Jest, Enzyme, etc., and how would have behaved differently if using a TDD approach? No need to implement it here of course, just start thinking about it.

PR #353, opened 2022-03-24
Ldn-Class -8 - Saf Shah - CYF Hotel - React - Week 1

s-shah-coder:


src/Footer.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <footer className="footer">
+      <ul>
+        {props.contact.map(string => (
```
Awyssa, line 7: In React, when we return jsx in a map, we need to give the first element a prop of key={ some number }. Some number can be literally any number, but one nice way we can do this is by using the index number of the array. So this will be 0,1,2,3,4... and so on. Not doing this dons't break React, but for in-memory and other weird javascript reasons, we should ALWAYS do this. If you look in your console, you will see this too.

The correct code is this below. But! Also note I used an implicit return map. By changing the {} to () the function knows to return the first thing, and therefore I don't have to write the keyword return. Neither way is better than the other, I just wanted to show you this incase you see it in the future.

```
<ul>
	{props.contact.map((string, index) => (
		<li key={index}>{string}</li>
	))}
        ;
</ul>
```

s-shah-coder: Thank you for the feedback, I will keep this in mind going forward!

PR #356, opened 2022-03-25
London class-8/Laleh Koupaei/React-Week1

lalehkoupa:
week1 is completed


lalehkoupa:
Hello @Awyssa ,my codes ready fr reviewing, many thanks

src/App.js
```diff
@@ -1,13 +1,50 @@
...
+    image:
+      "https://static.standard.co.uk/2021/10/26/10/2BD1M59-1.jpg?width=968&auto=webp&quality=50&crop=968%3A645%2Csmart",
+    website: "peoplemakeglasgow.com",
+    summary:
+      "Glasgow is a port city on the River Clyde in Scotland's western Lowlands. It's famed for its Victorian and art nouveau architecture, a rich legacy of the city's 18th–20th-century prosperity due to trade and shipbuilding. Today it's a national cultural hub, home to institutions including the Scottish Opera, Scottish Ballet and National Theatre of Scotland, as well as acclaimed museums and a thriving music scene."
+  }
+];
+
 const App = () => {
+  const footerDetail = [
```
Awyssa, line 37: This code is correct, however, I would personally move line 36 to line 9 and have all your let/const and javascript within the `App` function. This won't change any functionality, but the code will be much nicer to read :)

src/Footer.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <div>
+      <ul className="footer">
+        {props.footers.map(string => {
```
Awyssa, line 7: In React, when we return jsx in a map, we need to give the first element a prop of`key={ some number }`.   Some number can be literally any number, but one nice way we can do this is by using the index number of the array. So this will be 0,1,2,3,4... and so on. Not doing this dons't break React, but for in-memory and other weird javascript reasons we should do this. If you look in your console, you will see this too. 

The correct code is this below. But! Also note I used an implicit return map. By changing the `{}` to `()` the function knows to `return` the first thing, and therefore I don't have to write the keyword `return`. Neither way is better than the other, I just wanted to show you this incase you see it in the future.

```
<ul className="footer">
	{props.footers.map((string, index) => (
		<li key={index}>{string}</li>
	))}
</ul>
```

src/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
 
 const Bookings = () => {
   const search = searchVal => {
```
Awyssa, line 7: This is the best practise I think. Notice how all the code is wrapped in one function. We should always do this as it is much easier to read. When we start working with hundreds of components, using this method is very useful! So I would defiantly encourage you to start doing this now.

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import moment from "moment";
+
+const headings = [
```
Awyssa, line 4: Notice how this is outside of the SearchResults function? We should not do that. 
I would move line 15 `const SearchResults = props => {` to line 4, so we wrap all of our code in one function.

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
...
+  "Check-Out",
+  "Night"
+];
+const SearchResults = props => {
+  return (
+    <table className="table">
+      <thead>
+        <tr>
+          {headings.map(heading => {
+            return <th scope="col">{heading}</th>;
```
Awyssa, line 21: We are returning `jsx` in a map here. We need to give the first element a `key` prop. Look at other comment for details, or in your browser console.

The correct code is
```
<thead>
	<tr>
		{headings.map((heading, index) => {
			return <th key={index} scope="col">{heading}</th>;
		})}
	</tr>
</thead>
```

src/TouristInfoCards.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div className="tourist-card">
+      {props.citiesName.map(city => {
+        return (
+          <div className="card">
```
Awyssa, line 8: We are returning jsx in a map here. We need to give the first element a key prop. Look at other comment for details, or in your browser console.

The correct code is:
```
{props.citiesName.map((city, index) => {
	return (
	    <div key={index} className="card">
```

src/App.js
```diff
@@ -1,13 +1,50 @@
 import React from "react";
-
-import Bookings from "./Bookings";
+import Heading from "./Heading.js";
+import TouristInfoCards from "./TouristInfoCards.js";
+import Bookings from "./Bookings.js";
+import Footer from "./Footer.js";
+import SearchResults from "./SearchResults.js";
```
Awyssa, line 6: Error warning: `SearchResults` is defined but never used.

I imagine this one was hard to do! Let's speak about it in our next catch up :)

Awyssa:
Great job @lalehkoupa!!! For a first react project, this was never going to be easy... 

That being said, you smashed it!!! Really great job and you should be proud of what you have accomplished so far!

I added some comments, but they are mainly around coding standards. Your code works really well! 

Please look at my comments and we can go over them in our next catch up :D

lalehkoupa:
> Great job @lalehkoupa!!! For a first react project, this was never going to be easy...
> 
> That being said, you smashed it!!! Really great job and you should be proud of what you have accomplished so far!
> 
> I added some comments, but they are mainly around coding standards. This is just readability not functionality. Your code works perfectly, so don't worry about that!
> 
> Please look at my comments and we can go over them in our next catch up :D

lalehkoupa:
@Awyssa Thank you so much Michael for always giving lots of positive energy , and really appreciate for reviewing my work ,I will go through them and will apply all of the changes to my code.

src/TouristInfoCards.js
lalehkoupa: Thanks Done!

src/SearchResults.js
lalehkoupa: Done!

PR #360, opened 2022-03-25
London Class-8-Timea Reich-cyf-hotel-react

TimeaReich:


src/SearchResults.js
```diff
@@ -0,0 +1,39 @@
...
+import TableRow from "./TableRow";
+
+const SearchResult = (props) => {
+
+  const dataArray = props.data;
+  return (
+    <table className="table">
+      <thead className="thead-dark">
+        <tr>
+          <th scope="col">ID</th>
```
MatenRehimi, line 12: Could maybe have an array of column headers then use map to reduce the numbers of lines here

PR #361, opened 2022-03-25
London class 8 - Hamra Khan - React - Week1

HamraKhan:


src/Footer.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+function Footer(props) {
+  return (
+    <div className="footer">
+      <ul className="footer-info">
+        {props.contact.map(info => {
+          return (
+            <li className="info" key={info}>
```
ewintram, line 9: the key should always be unique, so it's recommended to use the index, e.g. `key={index}`. Can you remember how we can access the index when using `map`?

src/SearchResults.js
```diff
@@ -0,0 +1,51 @@
...
+  `numberOfNights`
+];
+
+function SearchResults(props) {
+  return (
+    <table className="table">
+      <thead className="table-header">
+        <tr>
+          {headings.map(heading => {
+            return <th scope="col">{heading}</th>;
```
ewintram, line 22: well done mapping the headings!

src/SearchResults.js
```diff
@@ -0,0 +1,51 @@
...
+          })}
+        </tr>
+      </thead>
+      <tbody>
+        {props.bookings.map(booking => {
+          return (
+            <tr>
```
ewintram, line 29: do you see any warnings in the console about missing keys..? If so, can you fix them using the index as the key?

PR #363, opened 2022-03-25
 London Class 8 - Hatice Aydogan - CYF-Hotel-React - Week 1

Hatice06:


src/Bookings.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
```
ewintram, line 4: it is convention to use Pascale case (where the first letter is also capitalised) only for JSX components, so this should be `import fakeBookings from "./data/fakeBookings.json";`

src/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <ul>
+      {props.contact.map((item, key) => (
+        <li key={key}>{item}</li>
```
ewintram, line 7: well done for including the key!

src/SearchResults.js
```diff
@@ -0,0 +1,41 @@
...
+            <th scope="row">{item.id}</th>
+            <td>{item.title}</td>
+            <td>{item.firstName}</td>
+            <td>{item.surname}</td>
+            <td>{item.email}</td>
+            <td>{item.roomId}</td>
+            <td>{item.checkInDate}</td>
+            <td>{item.checkOutDate}</td>
+            <td>
+              {moment(item.checkOutDate).diff(moment(item.checkInDate), "days")}
```
ewintram, line 32: well done for using moment

src/Bookings.js
Hatice06: Thanks Ellie I changed it.

PR #366, opened 2022-03-25
London Class 8 - Teniola Ogunjinrin- cyf-hotel-react

teniolao:


src/App.js
```diff
@@ -1,13 +1,26 @@
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Restaurant />
+      {/* //orders={orders} handleClick={incrementCount} */}
+      <Footer
+        details={[
```
MatenRehimi, line 18: Generally, we want our return to be as short to make it simpler to read. If the value for details was initialised above the return in a variable than doing details={exampleDetails} then your return would be shorter. Overall, same functionality before and after this change.

PR #369, opened 2022-03-25
London8-MandeepSangha-React-1

mandeepsangha:


src/SearchResults.js
```diff
@@ -0,0 +1,63 @@
...
+        setIsClicked(!isClicked);
+      }}
+      className={isClicked ? "highlighttrue" : ""}
+    >
+      <th scope="row">{booking.id}</th>
+      {headers.map(heading => {
+        if (heading === "id") {
+          return null;
+        } else if (heading === "#Nights") {
+          const inDate = moment(booking.checkOutDate);
```
MatenRehimi, line 52: Looks like inDate and outDate are backwards which still gives the correct diff so hard to spot

PR #381, opened 2022-03-31
London Class 8 - Hamra Khan - React - Week 2

HamraKhan:


src/Order.js
```diff
@@ -0,0 +1,18 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order({ orderType }) {
```
ewintram, line 4: well done for destructuring the props in the component definition

src/Order.js
```diff
@@ -0,0 +1,18 @@
...
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    setOrders(currentCount => currentCount + 1);
```
ewintram, line 8: you don't need to pass a function as an argument to the setOrders function here, because the state type is a number which is immutable (i.e. a new number will always be returned from a calculation). So this can be simplified to
```
function orderOne() {
    setOrders(orders + 1);
}
```

src/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function RestaurantButton(props) {
+  return (
+    <button onClick={props.orderOne} className="btn btn-primary">
```
ewintram, line 5: well done for passing the orderOne function as a prop

src/SearchResults.js
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function BookingItem(props) {
+  const booking = props.booking;
+  const index = props.key;
+  const [toggleSelected, setToggleSelected] = useState(false);
+  const [bookingItemStyle, setBookingItemStyle] = useState("default-row");
+
+  function changeColor() {
+    setToggleSelected(!toggleSelected);
```
ewintram, line 46: nice

PR #385, opened 2022-04-01
Cyf add

Sslim123:
i change some of function oreder to mahe it same clear and made some space between order and orderType


Sslim123:
hi Youssef can you review what i changed please

PR #392, opened 2022-04-07
London Class8-Marina Almeida-CYF-Hotel-React module week 2

MarinaAlmeida20:


src/components/search/HighlightRow.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from "react";
+import moment from "moment";
+
+function HighlightRow({ booking, i }) {
```
jackhclee, line 4: Align your code to use arrow function. See this link for the benefits of using arrow function. https://www.freecodecamp.org/news/when-and-why-you-should-use-es6-arrow-functions-and-when-you-shouldnt-3d851d7f0b26/

src/components/search/SearchButton.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+function SearchButton() {
```
jackhclee, line 3: Align your code to use arrow function. See this link for the benefits of using arrow function. https://www.freecodecamp.org/news/when-and-why-you-should-use-es6-arrow-functions-and-when-you-shouldnt-3d851d7f0b26/

PR #396, opened 2022-04-08
London Class 8 - Hatice Aydogan - CYF-Hotel-React - Week 3

Hatice06:


src/Search.js
```diff
@@ -1,23 +1,37 @@
...
+const Search = props => {
+  const [searchInput, setSearchInput] = useState("");
+
+  const handleSearchInput = event => {
+    setSearchInput(event.target.value);
+  };
+
+  const handleSubmit = event => {
+    event.preventDefault();
+    props.search(searchInput);
```
ewintram, line 13: well done using the prop function passed from the parent

src/SearchResults.js
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from "react";
+import moment from "moment";
+import CustomerProfile from "./CustomerProfile";
+
+const SearchResult = props => {
+  const [selectedId, setSelectedId] = useState();
+
+  const setSelected = id => {
```
ewintram, line 8: I don't think you need this function, as you can just pass the state hook to your `TableRow`, i.e.
```
<TableRow item={item} index={index} setSelected2={setSelectedId} />
```
Also, `setSelected2` is not a very clear name for a function

src/SearchResults.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+const TableRow = props => {
+  const [selected, setSelected] = useState(false);
+
+  const clickHandler = () => {
+    setSelected(!selected);
+  };
+
+  const handleCustomerId = id => {
```
ewintram, line 46: I also think you don't need this function, i.e. you can do this
```
<button
     className="btn btn-primary"
     onClick={() => {
        props.setSelected2(props.item.id);
     }}
>
```

src/SearchResults.js
Hatice06: Thank you Ellie, it makes sense, I changed it .

src/SearchResults.js
Hatice06: Thank you Ellie, it makes sense, I changed it .

PR #477, opened 2022-12-03
NW5 Manchester - Haroon Ali - React - Week 1

haroon-ali-dev:


src/components/SearchResults.js
```diff
@@ -0,0 +1,36 @@
...
+            <th scope="col">Check In Date</th>
+            <th scope="col">Check Out Date</th>
+            <th scope="col">Nights</th>
+            <th scope="col">Customer Profile</th>
+          </tr>
+        </thead>
+        <tbody>
+          {bookings.map((booking, i) => (
+            <SearchResult
+              key={i}
```
Dedekind561, line 26: Can you use another value like `booking` instead of `i` when setting the key prop inside this callback function?

src/components/TouristInfoCards.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+
+export default function TouristInfoCards() {
+  return (
+    <div className="cont-card">
+      <div className="card">
+        <img
```
Dedekind561, line 7: Looking good here! For further refactors, you have duplication of the divs with `className="card"` ( except for the actual info they display ). You can further abstract out this div into its own component and then render that component in here for every tourist info card you've got.

src/components/Bookings.js
```diff
@@ -0,0 +1,73 @@
...
+      } else {
+        setLoading(false);
+        setError(true);
+      }
+    };
+
+    getBookings();
+  }, []);
+
+  const search = searchVal => {
```
Dedekind561, line 30: I would rename this function - make it's name a bit more specific. What exactly is this function searching for? This might help you to come up with a more specific name.

src/components/Bookings.js
```diff
@@ -0,0 +1,73 @@
...
+      setCustomerProfile(data);
+    }
+  };
+
+  return (
+    <div className="App-content">
+      <div className="container-fluid">
+        <Search search={search} />
+        <div className="cont-search-results">
+          {loading ? <h1>Loading...</h1> : <></>}
```
Dedekind561, line 55:
```suggestion
          {loading ? <h1>Loading...</h1> : null}
```
Is rendering `null` preferable here? Not completely sure myself - just makes more sense to me to say render nothing in the case where the left hand condition evaluates to `false`. I might be wrong though.

src/components/CustomerProfile.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export default function CustomerProfile({ customerProfile }) {
+  return (
+    <div className="cont-profile">
```
Dedekind561, line 5: What does `"cont-profile"`  mean?

src/components/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+export default function Footer({ contactDetails }) {
+  return (
+    <footer>
+      <ul>
+        {contactDetails.map((detail, i) => (
+          <li key={i}>{detail}</li>
```
Dedekind561, line 8: Again best to avoid using `i` if possible for the key prop

src/components/Restaurant.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import Order from "./Order";
+
+const Restaurant = () => {
+  return (
+    <div className="cont-res">
```
Dedekind561, line 6: What does `cont-res` mean? Beware over shortened classNames like this - sometimes difficult to understand when you're reading code.

Dedekind561:
Hi @haroon-ali-dev,
Your code here is looking excellent! Very well organised, easy to read and clearly abstracted.
I've added a few comments for you to improve some things here and there - let me know if they make sense. 
Well done! :+1:

haroon-ali-dev:
> Hi @haroon-ali-dev, Your code here is looking excellent! Very well organised, easy to read and clearly abstracted. I've added a few comments for you to improve some things here and there - let me know if they make sense. Well done! +1

thank you Mitch. all makes sense and is very helpful :+1:

src/components/Bookings.js
salehpirouzi-cyf: how about this instead?
```suggestion
          {loading && <h1>Loading...</h1>}
```

PR #483, opened 2022-12-04
NW5-MANCHESTER -ADIBA BELAYETE- REACT-WEEK_01-CYF HOTEL PROJECT

Adibab:


src/App.js
```diff
@@ -6,10 +6,95 @@ import "./App.css";
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      {/* task 02 */}
+      <Heading />
+      {/* task 3, */}
+      <TouristInfoCards />
       <Bookings />
+      {/* task 4 */}
+      <Footer address={address} />
```
Nomes27, line 15: I like how you have used comments to break down the different tasks. Good use of components here

src/App.js
```diff
@@ -6,10 +6,95 @@ import "./App.css";
...
+    <header className="App-header">
+      <img
+        className=".App-logo "
+        src="https://image.flaticon.com/icons/svg/139/139899.svg"
+        alt="hotel's logo"
+      />
+      CYF Hotel
+    </header>
+  );
+}
```
Nomes27, line 31: this is a good component well done, but extract it out into it's own file & import it into the `App` component. Same with your `TouristInfoCards` component and `Footer` component. This helps with reusability, for example with your `Heading` component at the moment it can only be accessed within `App`, but if you make it it's own component you can import this anywhere you like in your app and use it (you can imagine this would be especially useful with a large scale app)

src/App.js
```diff
@@ -6,10 +6,95 @@ import "./App.css";
...
+              <h5 className="card-title fw-bolder">{city.name} </h5>
+              <p class="card-text"> {city.description}</p>
+              <a href={city.link} className="btn btn-primary">
+                Go somewhere
+              </a>
+            </div>
+          </div>
+        );
+      })}
+    </div>
```
Nomes27, line 81: nice use of .map here to render each card, the styling looks really good too

src/Bookings.js
```diff
@@ -12,7 +12,8 @@ const Bookings = () => {
     <div className="App-content">
       <div className="container">
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
```
Nomes27, line 15: this is good that you're passing in your data as props to use in the `SearchResults` component

src/Search.js
```diff
@@ -26,4 +27,8 @@ const Search = () => {
   );
 };
 
+function SearchButton() {
+  return <button className="btn btn-primary">Search</button>;
+}
+
```
Nomes27, line 33: like mentioned above, extract this out into it's own component :) then anywhere in your app you needed a search button you could import it in

src/SearchResults.js
```diff
@@ -0,0 +1,53 @@
...
+          <tr>
+            <th scope="col">id</th>
+            <th scope="col">Title</th>
+            <th scope="col">First Name</th>
+            <th scope="col">SurName</th>
+            <th scope="col">Email</th>
+            <th scope="col">Room ID</th>
+            <th scope="col">check in date</th>
+            <th scope="col">check out date</th>
+            <th scope="col"> Number of Nights </th>
```
Nomes27, line 20: you could even have a think about a dynamic way to get the table headings here, so you don't have to keep updating it if the data your received changed (for example imagine if more columns were added, or some were removed). You get each of these keys with each object, so have a think how you could grab the keys without having to type them all out

src/SearchResults.js
```diff
@@ -0,0 +1,53 @@
...
+                <td> {profile.roomId}</td>
+                <td> {profile.checkInDate}</td>
+                <td> {profile.checkOutDate}</td>
+                <td>
+                  {" "}
+                  {Math.abs(
+                    Moment(profile.checkOutDate).diff(
+                      Moment(profile.checkInDate),
+                      "days"
+                    )
```
Nomes27, line 41: good use of moment here to get the total number of nights

Nomes27 (commented):
really good start Adiba, well done :) just remember to extract the component out into it's own file and export it (and then import it wherever you need to use it)

src/Bookings.js
```diff
@@ -1,18 +1,46 @@
...
+    // console.info("TO DO!", searchVal);
+    const newBookings = bookings.filter(booking => {
+      return (
+        booking.firstName.toLowerCase() === searchVal.toLowerCase() ||
+        booking.surname.toLowerCase() === searchVal.toLowerCase()
+      );
+    });
+
+    setBookings(newBookings);
   };
```
Nomes27, line 29: Using the filter method here works really well, good job using `.toLowerCase() ` too to check for both upper and lower case matches :)
To improve this search function, think about how you might make it work if you were to make multiple searches for people - for example, at the moment if you search for 'john' and then search for 'judi' afterwards it doesn't bring up judi's profile (because the filtered bookings from when you searched for John are now saved in state.) Can you think of a way around this?

src/Button.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+//  task-11
+function RestaurantButton(props) {
+  return (
+    <button className="btn btn-primary" onClick={props.handleClick}>
+      Add
+    </button>
+  );
```
Nomes27, line 8: this is good, you're passing your handleClick function to your child component to use, this means you can update the state in the parent component

src/CustomerProfile.js
```diff
@@ -0,0 +1,19 @@
...
+      {customerProfile.length !== 0 ? (
+        <ul>
+          <li>Id: {customerProfile.id}</li>
+          <li>Email: {customerProfile.email}</li>
+          <li>Vip: {customerProfile.vip ? "Yes" : "No"}</li>
+          <li>Phone no: {customerProfile.phoneNumber}</li>
+        </ul>
+      ) : (
+        <></>
+      )}
```
Nomes27, line 16: a ternary operator works here, but you can also use `&&` in react, for example:
```
 {customerProfile.length !== 0 && (
 //code you want to display if the statement is true
```
this is handy if you don't want to render something if the statement is false

src/Order.js
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <div>
+      <li>
+        {props.orderType}: {Orders}
+        {/* task - 11 */}
+        <RestaurantButton handleClick={orderOne} />
+      </li>
+    </div>
+  );
+}
```
Nomes27, line 23: this is good, you're rendering an order component for each of the different food options & on button click, the order count increases independently for each food type, nice job :)

src/Order.js
```diff
@@ -0,0 +1,25 @@
...
+import RestaurantButton from "./Button";
+
+//  task -12
+function Order(props) {
+  // task -09
+  const [Orders, setOrders] = useState(0);
+
+  // task-10
+  function orderOne() {
+    setOrders(count => count + 1);
```
Nomes27, line 11: here you can just do `setOrders(count+1)`

src/Search.js
```diff
@@ -1,23 +1,42 @@
...
-const Search = () => {
   return (
     <div className="search">
       <div className="page-header">
         <h4 className="text-left">Search Bookings</h4>
       </div>
       <div className="row search-wrapper">
         <div className="col">
-          <form className="form-group search-box">
+          <form className="form-group search-box" onClick={onSubmit}>
```
Nomes27, line 23: for forms we use `onSubmit` rather than `onClick`

src/Search.js
```diff
@@ -1,23 +1,42 @@
...
             <label htmlFor="customerName">Customer name</label>
             <div className="search-row">
               <input
                 type="text"
                 id="customerName"
                 className="form-control"
                 placeholder="Customer name"
+                value={searchInput}
+                onChange={handleSearchInput}
```
Nomes27, line 32: good that you're using `htmlFor` and `id` to link the label & input. Your handleSearchInput function works well

src/Tablerow.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState } from "react";
+import Moment from "moment";
+
+function CreateTableRow({ profile, index, showProfile }) {
```
Nomes27, line 4: creating a new component for each table row works really well, this means you can easily control the state of each row when deciding if the row should be highlighted or not

src/Tablerow.js
```diff
@@ -0,0 +1,43 @@
...
+  const [colour, setColor] = useState(false);
+  function handleClick() {
+    setColor(oldColour => !oldColour);
+  }
```
Nomes27, line 8: you already have the colour variable in state, so you can just have this as `setColour(!colour)`, that way it sets the colour boolean to the opposite of what is currently in state

PR #486, opened 2022-12-05
NW5 Manchester - Doris Siu - React - Week 1

Doris-Siu:


src/App.js
```diff
@@ -1,13 +1,23 @@
...
+    "0123 456789"
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      {/* <header className="App-header">CYF Hotel</header> */}
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Footer address={address} />
```
Nomes27, line 20: good use of components here, you're importing them & passing the address as props to use in your footer component, well done :)

src/Search.js
```diff
@@ -17,7 +18,8 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
+              {/* <button className="btn btn-primary">Search</button> */}
+              <SearchButton />
```
Nomes27, line 22: this is good, you've extracted the button code into a new SearchButton component and imported it, well done

src/SearchResults.js
```diff
@@ -0,0 +1,65 @@
...
+          <tr>
+            <th scope="col">Title</th>
+            <th scope="col">First name</th>
+            <th scope="col">Surname</th>
+            <th scope="col">Email</th>
+            <th scope="col">Room ID</th>
+            <th scope="col">Check-in Date</th>
+            <th scope="col">Check-out Date</th>
+            <th scope="col">Number of nights</th>
+          </tr>
```
Nomes27, line 41: you could have a think about a dynamic way of grabbing the column headers (see if the object keys may help with this) so you don't have to type them out/update them if the data you are receiving changes

src/SearchResults.js
```diff
@@ -0,0 +1,65 @@
...
+              <td>{booking.title}</td>
+              <td>{booking.firstName}</td>
+              <td>{booking.surname}</td>
+              <td>{booking.email}</td>
+              <td>{booking.roomId}</td>
+              <td>{booking.checkInDate}</td>
+              <td>{booking.checkOutDate}</td>
+              <td>
+                {Moment(booking.checkOutDate).diff(
+                  Moment(booking.checkInDate),
```
Nomes27, line 55: nice use of moment here to show the total nights booked in

src/TouristInfoCards.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import card from "./data/card";
+
+export default function TouristInfoCards() {
+  const cardInfo = card();
```
Nomes27, line 5: this works, but you could also just have your data in your `card` file saved a variable and export it, rather than needing to call a function to get your data

src/TouristInfoCards.js
```diff
@@ -0,0 +1,22 @@
...
+          <img src={element.image} alt="" className="card-img-top" />
+          <div className="card-body">
+            <p>{element.cityName}</p>
+            <p>{element.text}</p>
+            <a href={element.link} className="btn btn-primary">
+              More information
+            </a>
+          </div>
+        </span>
+      ))}
```
Nomes27, line 19: this is good, you're rendering the info for each card and everything is displayed nicely in the browser

Nomes27 (commented):
This is a great start - you've extracted code out into components and imported them successfully, well done. Just remember to add keys when you're mapping in react - I know it can be easy to forget sometimes, but there should be warnings in the console if you miss any!

src/Bookings.js
```diff
@@ -1,18 +1,45 @@
...
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    let filteredBookings = [];
+    bookings.forEach(booking => {
+      if (
+        booking.firstName.toLowerCase().includes(searchVal.toLowerCase()) ||
+        booking.surname.toLowerCase().includes(searchVal.toLowerCase())
+      ) {
+        filteredBookings.push(booking);
+      }
```
Nomes27, line 17: good that you're converting both the search val and booking to lowercase here :). `forEach` works, but as you are filtering the bookings, the `filter` method would work perfectly for this kind of task

src/Bookings.js
```diff
@@ -1,18 +1,45 @@
...
+    });
+    setBookings(filteredBookings);
   };
 
+  const [bookings, setBookings] = useState(0);
+  const [profile, setProfile] = useState(0);
```
Nomes27, line 23: try and declare your state at the top of the component

src/Bookings.js
```diff
@@ -1,18 +1,45 @@
...
+    let filteredBookings = [];
+    bookings.forEach(booking => {
+      if (
+        booking.firstName.toLowerCase().includes(searchVal.toLowerCase()) ||
+        booking.surname.toLowerCase().includes(searchVal.toLowerCase())
+      ) {
+        filteredBookings.push(booking);
+      }
+    });
+    setBookings(filteredBookings);
```
Nomes27, line 19: your search functionality works well, well done. An improvement could be to consider the following:
if you search for one person, for example 'john' and then search for 'judi' it doesn't return anything, as your bookings state has already been set to the filtered results for 'john' (which doesn't include judi) - can you have a think how you might solve this?

src/Bookings.js
```diff
@@ -1,18 +1,45 @@
...
   };
 
+  const [bookings, setBookings] = useState(0);
+  const [profile, setProfile] = useState(0);
+
+  useEffect(() => {
+    console.log("use effect...");
+    // fetch("https://cyf-react.glitch.me/delayed")
+    fetch("https://cyf-react.glitch.me/error")
```
Nomes27, line 28: assuming you must be testing out the error response and that's why this is in here :)

src/Bookings.js
Nomes27, line 23: rather than setting your state as 0 for no data, try and use `null` or `[]` instead, I noticed you're using `profile` for the customerId, but if a customer had an id of 0 this could cause problems down the line

src/CustomerProfile.js
```diff
@@ -0,0 +1,23 @@
+import React, { useState, useEffect } from "react";
+
+export default function CustomerProfile({ customerId }) {
+  const [customerInfo, setCustomerInfo] = useState({});
+  useEffect(() => {
+    console.log("...");
+    fetch(`https://cyf-react.glitch.me/customers/${customerId}`)
+      .then(res => res.json())
+      .then(data => setCustomerInfo(data));
+  }, [customerId]);
```
Nomes27, line 10: good, you've added the id to the dependency array so that a new request is made whenever it changes, well done :)

src/CustomerProfile.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <div>
+      {customerId === 0 ? null : (
```
Nomes27, line 13: if you are just wanting to display something when a statement equates to true, you can use `&&` in react, for example:
```
{customerId && (
//content you want to render
```
mentioned above but I would use null for your customerId rather than 0, setting as 0 is a bit confusing & can lead to problems if a customer does have an id of 0

src/CustomerProfile.js
```diff
@@ -0,0 +1,23 @@
...
+        <>
+          <p>Customer ID: {customerInfo.id}</p>
+          <p>Email: {customerInfo.email}</p>
+          <p>VIP: {customerInfo.vip ? "yes" : "no"}</p>
+          <p>Phone number: {customerInfo.phoneNumber}</p>
+        </>
+      )}
+    </div>
+  );
+}
```
Nomes27, line 23: your customer profile component displays nicely on show profile button click, well done

src/Order.js
```diff
@@ -0,0 +1,14 @@
...
+  const [orders, setOrders] = useState(0);
+  function orderOne() {
+    return setOrders(orders + 1);
+  }
+  return (
+    <li>
+      {orderType}: {orders} <RestaurantButton handleClick={orderOne} />
+    </li>
+  );
+}
```
Nomes27, line 14: this is good, your `orderOne` function works well & your order count increases independently for each food item :)

src/Restaurant.js
```diff
@@ -1,14 +1,25 @@
-import React from "react";
+import React, { useState } from "react";
+import Order from "./Order";
 
 const Restaurant = () => {
-  const pizzas = 0;
+  const [orders, setOrders] = useState(0);
+  function orderOne() {
+    return setOrders(orders + 1);
+  }
```
Nomes27, line 8: remember to remove unnecessary state & functions when they're no longer being used, same for commented out code

src/Search.js
```diff
@@ -1,23 +1,43 @@
...
+import React, { useState } from "react";
+import SearchButton from "./SearchButton";
+
+const Search = ({ search }) => {
+  const [searchInput, setsearchInput] = useState("");
+
+  function handleSearchInput(e) {
+    setsearchInput(e.target.value);
+    //console.log(e.target.value);
+  }
```
Nomes27, line 10: this works well to handle the search & nice clear naming of the function too

src/SearchResults.js
```diff
@@ -0,0 +1,101 @@
...
+  //     title: "Doctor",
+  //     firstName: "Sadia",
+  //     surname: "Begum",
+  //     email: "begum_sadia@sadia.org",
+  //     roomId: 1,
+  //     checkInDate: "2018-02-15",
+  //     checkOutDate: "2018-02-28"
+  //   }
+  // ];
+  let bookings = results;
```
Nomes27, line 28: do you need to declare a new variable here?

src/SearchResults.js
```diff
@@ -0,0 +1,101 @@
...
+        let updated = { ...bookingColor };
+        updated[bookingId] = "white";
+        return updated;
+      } else {
+        let updated = { ...bookingColor };
+        updated[bookingId] = "yellow";
+        return updated;
+      }
+    });
+  };
```
Nomes27, line 44: good job on getting each row to highlight on click. A simpler way to achieve this could be to create a component for each row, each row can then have its own state, which can be toggled on click (for example a highlighted state value could be toggled between true and false, with a style/className being set depending on whether it is true or false)

src/SearchResults.js
```diff
@@ -0,0 +1,101 @@
...
+                  >
+                    Show profile
+                  </button>
+                </td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      ) : (
+        <h1>Please wait. The booking data are loading...</h1>
```
Nomes27, line 97: good that you're thinking about what to display to a user while their data is loading

PR #488, opened 2022-12-05
NW5-Manchester-Zaw Khing-React-Week-1

zkhing:


src/SearchResults.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import Moment from "react-moment";
+import TableRow from "./TableRow";
+
+const SearchResults = props => {
+  console.log(props);
```
Dedekind561, line 6: Hey @zkhing, be sure to remove any rogue `console.log`s from your code - they should only be their for debugging purposes. Your production code shouldn't contain any logs so it's good to get into the habit early on.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+
+const TouristInfoCards = () => {
+  return (
+    <div className="flex-container">
+      {/* About Glasgow */}
```
Dedekind561, line 6: :eyes: Comment here

src/TableRow.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import Moment from "react-moment";
+import { useState } from "react";
+
+const TableRow = ({
+  id,
```
Dedekind561, line 6: Nice use of parameter destructuring here! :+1:

src/Order.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { useState } from "react";
+
+const Order = ({ orderType }) => {
+  //   const orderType = props.orderType;
```
Dedekind561, line 5: Make sure you remove any rogue comments like this - don't think you need this one anymore

src/App.css
```diff
@@ -39,7 +44,7 @@
   padding: 5px 0 20px 0;
 }
 
-tr {
+.btn btn-primary tr {
```
savagematt, line 47: See: https://www.w3schools.com/cssref/css_selectors.php

This selector applies styles to:

* all `tr` elements...
* inside a `btn-primary` element (which doesn't exist- `btn-primary` is a css class)...
* inside an element with `class=btn`

I'm not sure what your intention was here- let's chat about it when we meet.

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
+  }
+
   const search = searchVal => {
+    const filteredName = bookings.filter(
+      booking =>
+        booking.firstName.toLowerCase() === searchVal ||
+        booking.surname.toLowerCase() === searchVal
+    );
+    setBookings(filteredName);
     console.info("TO DO!", searchVal);
```
savagematt, line 30: It's best to avoid committing any console statements. 

They are sometimes ok for debugging as long as you only use them locally and remove them before pushing your changes.

If you don't keep on top of use of `console`, as the codebase grows you accumulate more and more of them and it can get messy and confusing, and annoying for other developers working on the codebase.

savagematt: Reminder for next checkin- I can show you how to use the Chrome debug tools so you don't need to do this.

src/TableRow.js
```diff
@@ -0,0 +1,55 @@
...
+  surname,
+  email,
+  roomId,
+  checkInDate,
+  checkOutDate,
+  showProfile
+}) => {
+  const [color, setColor] = useState(false);
+
+  function getProfile() {
```
savagematt, line 18: This is not essential, but when you click the "Show Profile" button, the row also gets highlighted. That could be confusing to users.

This happens because the click event gets handled by `button`, but then also bubbles up to the `tr` that the `button` is inside, and the `onClick` on `tr` highlights the row.

If you wanted the "Show Profile" button to just display the profile `div`, without  also highlighting the row, you could call `stopPropagation()`, so the event is handled only by `button` and then React prevents it reaching parent components. 

Like this:

```
function getProfile(e) {
   e.stopPropagation();
   //... rest of your code
```

For more info, see: https://reactjs.org/docs/events.html#overview

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
+
+  function showProfile(profile) {
+    setProfile(profile);
+  }
+
   const search = searchVal => {
+    const filteredName = bookings.filter(
+      booking =>
+        booking.firstName.toLowerCase() === searchVal ||
+        booking.surname.toLowerCase() === searchVal
```
savagematt, line 27: Try searching for "JOHN" instead of "john"- what happens? Can you see why?

savagematt: See also line 37

src/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <div className="footer">
+      <ul>
+        {props.contact.map((address, index) => (
+          <li key={index}> {address} </li>
```
savagematt, line 8: Remembered `key`- nice!

src/Heading.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <div>
+      <header className="App-header">CYF Hotel</header>
+      <img
+        className="hotel-logo"
+        src="https://cdn.dribbble.com/users/4031439/screenshots/10789004/media/4b07136e303947c2f2075b5b1995cf20.jpg?compress=1&resize=400x300&vertical=top"
+        alt=""
```
savagematt, line 10: It's good to get in to the habit of filling in `alt` for users on screen readers

src/Search.js
```diff
@@ -1,23 +1,38 @@
-import React from "react";
+import React, { useState } from "react";
+import SearchButton from "./SearchButton";
+
+const Search = ({ search }) => {
+  const [searchInput, setSearchInput] = useState("");
+
+  function handleSearchInput(event) {
+    setSearchInput(event.target.value);
```
savagematt, line 8: What happens if I delete the text in the search box and press "Search" again?

src/SearchButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const SearchButton = () => {
+  return (
+    <div>
+      <button className="btn btn-primary">Search</button>
```
savagematt, line 6: I _think_ this should also have `type="submit"`, because it's intended to submit the enclosing form.

Not essential because for some reason the code still works without it- I think React is doing something clever here to make it work as you'd expect.

savagematt (commented):
Zaw, this is looking really, really good. You've grasped how hooks work perfectly, and that's one of the hardest bits of React for most people.

PR #489, opened 2022-12-06
NW5-Manchester-Refik Gulak-cyf-hotel-react week-1

Refik48:


src/CustomerProfile.JS
```diff
@@ -0,0 +1,20 @@
+import React, { useState, useEffect } from "react";
```
savagematt, line 1: Need to rename the file with a lowercase .js instead of .JS

src/Heading.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <header className="App-header">
+      <img className="logo" alt="Heading logo" src="../public/Hotel1.png" />
```
savagematt, line 6: create-react-app does some magic with the `public` directory- the correct `src` for this is just `/Hotel1.png`

src/SearchResults.js
```diff
@@ -0,0 +1,45 @@
...
+            <th scope="col">E-mail</th>
+            <th scope="col">Room No</th>
+            <th scope="col">Check-in Date</th>
+            <th scope="col">Check-out Date</th>
+            <th scope="col">Numer of Nights</th>
+            <th scope="col">Profile</th>
+          </tr>
+        </thead>
+        <tbody>
+          {props.results.map((booking, index) => {
```
savagematt, line 29: If you run the app with `npm run start`, and then open [Chrome developer tools ](https://developer.chrome.com/docs/devtools/open/), then go to the console tab you'll see it complaining that `props` isn't defined.

This is because you've already destructured `{results}` on line 4.

So you can just say `{results.map...` here.

src/Bookings.js
```diff
@@ -1,20 +1,39 @@
...
+import Search from "./Search";
+import SearchResults from "./SearchResults";
 
 const Bookings = () => {
+  const [bookings, setBookings] = useState(null);
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
+      .then(res => res.json())
+      .then(data => {
+        console.log(data);
```
savagematt, line 11: Generally you want to avoid using `console.log` in project where it's not just you working on it, because it tends to clutter up the console output.

I'll show you how to use the debugger in Chrome developer tools so you can get the same information without using `console.log`

src/Bookings.js
```diff
@@ -1,20 +1,39 @@
...
+      });
+  }, []);
+
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    const selectedBooking = bookings.filter(
+      item =>
+        item.firstName.toLowerCase().includes(searchVal.toLowerCase()) ||
+        item.surname.toLowerCase().includes(searchVal.toLowerCase())
+    );
```
savagematt, line 22: This is really good! It's cool that you've lowercased _both_ the search terms and the fields you're searching over.

I also like that you're using `includes` instead of `===`, to allow users to search for any part of the string.

There could be a good argument for using `startsWith` as an alternative to `includes`, because this might reduce unwanted results in the search output. But this is the sort of thing you'd usually discuss with the customer, or your business analyst.

src/CustomerProfile.JS
savagematt: You can get this feedback by typing `npm run start` into the terminal.

src/CustomerProfile.JS
```diff
@@ -0,0 +1,20 @@
...
+
+const CustomerProfile = (props) => {
+  console.log(props);
```
savagematt, line 4: See earlier comments on using `console.log`

src/CustomerProfile.JS
```diff
@@ -0,0 +1,20 @@
...
+  const [profileId, setProfileId] = useState("");
+
+  useEffect(() => {
+    setProfileId(props.id);
+  }, [props.id]);
+
+  console.log(props.id);
```
savagematt, line 11: You don't need an effect at this point in the exercise, because we're just displaying `props.id` rather than fetching data from somewhere else. But it's fine to have this here as a placeholder.

`useEffect` will make more sense in [Lesson 3 point 21](https://github.com/CodeYourFuture/cyf-hotel-react#21-display-a-customer-profile---step-2)

src/Search.js
```diff
@@ -1,23 +1,37 @@
-import React from "react";
+import React, { useState } from "react";
+import SearchButton from "./SearchButton";
 
-const Search = () => {
+const Search = props => {
+  const [searchInput, setSearchInput] = useState("");
+  function handleSearchInput(event) {
+    console.log(event.target.value);
```
savagematt, line 7: See comments on `console.log` above

src/SearchTableRow.js
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <tr className={isSelected ? "bg-info" : null} onClick={handleClick}>
+      <td id="id">{booking.id}</td>
+      <td id="title">{booking.title}</td>
+      <td id="fName">{booking.firstName}</td>
+      <td id="sName">{booking.surname}</td>
+      <td id="email">{booking.email}</td>
+      <td id="roomNo">{booking.roomId}</td>
+      <td id="chekIn">{booking.checkInDate}</td>
+      <td id="checkOut">{booking.checkOutDate}</td>
```
savagematt, line 21: Nice that you add `id` here- this makes it easier to find these columns during testing later on

src/TouristInfoCards.js
```diff
@@ -0,0 +1,23 @@
...
+  return props.cardArr.map((card, index) => {
+    return (
+      <div key={index} className="card m-5" style={{ width: 18 + "rem" }}>
+        <img src={card.img} alt="City views " className="card-img-top" />
+        <div className="card-body">
+          <p>{card.text}</p>
+          <a
+            href={card.link}
+            rel="noopener noreferrer"
+            target="_blank"
```
savagematt, line 12: Good that you've thought about opening off-site links in a new tab

PR #492, opened 2022-12-06
NW5 Manchester - Luan Souza - React - Week 1

luan-murilo-de-oliveira-e-souza:


Dominic-Taylor-Dev (commented):
I really like what you did here. A few things I spotted in terms of user experience:

- Search works to some extent, but once I've searched I can't 'unsearch' to see the full list again. It would be nice if the search was 'live' i.e. when I delete my search term, I see the full list, as I type letters, the list gets smaller. As a minimum, I should be able to get result for another search if I type in a new word
- After adding a new booking, when I click the button the popup should disappear
- When I create a new booking, and then 'show profile', all of the fields like customer ID are empty. Even things like email that I just filled out. This seems a little odd

I'll have a look at the code now and see if anything jumps out

PR #494, opened 2022-12-06
nw5/shimen-afshar/react/week1

ShimenAfshar:


src/App.js
```diff
@@ -1,13 +1,26 @@
...
+// import SearchButton from "./SearchButton";
+import TouristInfoCards from "./TouristInfoCards";
+import Footer from "./Footer";
+// import SearchResults from "./SearchResults.JS";
+
+const data = [
+  "123 Fake Street, London, E1 4UD",
+  "hello@fakehotel.com",
+  "0123 456789"
+];
```
Nomes27, line 14: I would try and give this more of an explicit variable name, something like addressData maybe

src/App.js
```diff
@@ -1,13 +1,26 @@
...
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      {/* <header className="App-header">CYF Hotel</header> */}
+      <Heading />
       <Bookings />
+      <TouristInfoCards />
+      <Footer data={data} />
```
Nomes27, line 23: this is good, you've imported your components and you're passing your address data as props to your `Footer` component

src/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+function Footer({ data }) {
```
Nomes27, line 3: nice destructuring here :)

src/Heading.js
```diff
@@ -0,0 +1,17 @@
...
+      <img
+        src="https://i.pinimg.com/736x/10/ff/aa/10ffaadab6bc3c4c1dd4a3e44bf6d5ad.jpg"
+        alt=""
+        width="40"
+        height="25"
+      />
+      CYF Hotel
+    </header>
+  );
+};
```
Nomes27, line 15: this is good, you've extracted out the header code into it's own component. This means the component is now reusable can be used throughout the app (imagine you wanted lots of different headers, you could use the same `Header` component and just pass in different data)

src/SearchResults.JS
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from "react";
```
Nomes27, line 1: not sure what you're trying to import here?

src/SearchResults.JS
```diff
@@ -0,0 +1,48 @@
...
+      <td>{item.roomId}</td>
+      <td>{item.checkInDate}</td>
+      <td>{item.checkOutDate}</td>
+      <td>{checkDiff(item.checkInDate,item.checkOutDate)}</td>
+    </tr>
+            )
+        }
+        )
+    }
+  </tbody>
```
Nomes27, line 38: your formatting is a bit random in this file, make sure you have a code formatter enabled like prettier, so your code is nice & readable

src/SearchResults.JS
```diff
@@ -0,0 +1,48 @@
...
+</table>
+    </div>
+}
+let checkDiff = (InDate, OutDate) => {
+    const checkInDate = Moment(InDate, "YYYY-MM-DD");
+    const checkOutDate = Moment(OutDate, "YYYY-MM-DD");
+    const diff = checkOutDate.diff(checkInDate, "days");
+    return diff;
+  };
+export default SearchResult;
```
Nomes27, line 48: think you accidentally capitalised the js when naming this file, it wasn't loading for me initially

src/SearchResults.JS
```diff
@@ -0,0 +1,48 @@
...
+            return(
+                <tr>
+      <th scope="row">{item.id}</th>
+      <td>{item.title}</td>
+      <td>{item.firstName}</td>
+      <td>{item.sureName}</td>
+      <td>{item.email}</td>
+      <td>{item.roomId}</td>
+      <td>{item.checkInDate}</td>
+      <td>{item.checkOutDate}</td>
```
Nomes27, line 31: this is good, your data is showing up as expected in the table & you've used moment to calculate the nights stayed, well done

src/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
...
+        link="https://visitlondon.com"
+        image="https://media.istockphoto.com/id/1294454411/photo/london-symbols-with-big-ben-double-decker-buses-and-red-phone-booth-in-england-uk.jpg?b=1&s=170667a&w=0&k=20&c=SR5FBt8VvpZEOIxZcTmoEf9XxU0oiSnbobD3E6ZB358="
+        name="London"
+        text="Welcome to London! Discover the best of London with Visit London."
+      />
+    </div>
+  );
+};
+
+const Card = props => {
```
Nomes27, line 30: using a reusable `Card` component here is a really good idea, as you're essentially going to be using the same card layout for each city. I would pull this out into it's own file, rather than having it at the bottom here (and import it at the top). Having it in it's own file means it can be used throughout the app if needed and not just in `TouristInfoCards`.

You could also get rid of some repetition in the top section - for example if you have an array of data for all the cities, you could map over this, rendering a `Card` component for each. Imagine if you wanted to add 10 more city cards, your file would soon get pretty huge, so using a .map would help with this

Nomes27 (commented):
Well done Shimen, this is a really good start, you've created new reusable components and passed data to these through props :)

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
-    console.info("TO DO!", searchVal);
+    // console.info("TO DO!", searchVal);
+    searchVal = searchVal.toLowerCase();
+    const filterd = booking.filter(
+      item =>
+        item.firstName.toLowerCase() === searchVal ||
+        item.surname.toLowerCase() === searchVal
+    );
+    setBooking(filterd);
+  };
```
Nomes27, line 27: good, your filter function works well & good consideration of case too :) Something to consider for an improvement, if I search one name, 'john', followed by another 'judi', I don't get any results showing up (as the filtered list for john is in state when I search for judi) - can you have at think of a way this could be solved?

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
+
+  const showProfile = async id => {
```
Nomes27, line 29: good exploration of async await here

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
+    const res = await fetch(`https://cyf-react.glitch.me/customers/${id}`);
+    if (res.status === 200) {
+      const data = await res.json();
+      setCustomerProfile(data);
+    }
```
Nomes27, line 34: can you think of a way this request could make use of `useEffect`? for example, if we had some state tracking when the show profile button was clicked, we could use the dependency array to make the request whenever this state changed

src/CustomerProfile.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export default function CustomerProfile({ customerProfile }) {
```
Nomes27, line 3: your customer profile displays nicely on show profile button click, nice job

src/CustomerProfile.js
```diff
@@ -0,0 +1,19 @@
...
+      {customerProfile.length !== 0 ? (
+        <ul>
+          <li>Id: {customerProfile.id}</li>
+          <li>Email: {customerProfile.email}</li>
+          <li>Vip: {customerProfile.vip ? "Yes" : "No"}</li>
+          <li>Phone no: {customerProfile.phoneNumber}</li>
+        </ul>
+      ) : (
+        <></>
+      )}
```
Nomes27, line 16: here instead of a ternary operator, we can use `&&` (useful for when we don't care about rendering anything for the false condition) for example:
```
{customerProfile.length > 0 && (
//code you want to render
```

src/Order.js
```diff
@@ -0,0 +1,17 @@
...
+  }
+  return (
+    <>
+      <li style={{ margin: "20px" }}>
+        {orderType}:{order}
+        <RestaurantButton orderOneProps={orderOne} />
+      </li>
+    </>
+  );
+}
```
Nomes27, line 17: this component works well, by making a reusable component of `Order` you can apply the same logic & layout for multiple food items & each food item's count can be increased independently :)

src/Restaurant.js
```diff
@@ -1,14 +1,20 @@
 import React from "react";
+// import RestaurantButton from "./RestaurantButton";
+import Order from "./Order";
 
 const Restaurant = () => {
-  const pizzas = 0;
+  // const pizzas = 0;
```
Nomes27, line 6: remember to remove commented out code, it helps keep your files readable

src/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export default function RestaurantButton({ orderOneProps }) {
+  return (
+    <div>
+      <button className="btn btn-primary" onClick={orderOneProps}>
```
Nomes27, line 6: this is good, you're passing your `orderOne` function into your child component, so that when it's called it can update the state in the parent component (this is something we see a lot of in react)

src/Search.js
```diff
@@ -1,23 +1,36 @@
...
+          <form className="form-group search-box" onSubmit={submitHandle}>
             <label htmlFor="customerName">Customer name</label>
             <div className="search-row">
               <input
                 type="text"
                 id="customerName"
                 className="form-control"
                 placeholder="Customer name"
+                value={searchInput}
+                onChange={handleSearchInput}
```
Nomes27, line 31: this is a good `onChange` function & updates your state with the search input

src/SearchResult.js
```diff
@@ -0,0 +1,29 @@
+import React, { useState } from "react";
+
+export default function SearchResult({ item, checkDiff, showProfile }) {
+  const [select, setSelect] = useState(false);
+
+  function changeSelected() {
+    setSelect(!select);
+  }
```
Nomes27, line 8: breaking down each row into its own `SearchResult` component works really well, this means you can easily highlight each row, due to each row having its own state

PR #497, opened 2022-12-06
NW5-MANCEHSTER-Miguel-Cabral-React-WEEK1

Miguel-Cabral:


src/App.css
```diff
@@ -54,4 +53,16 @@ tr {
...
+}
+
+.logo {
+  width: 40px;
+}
+
+.cardsEl {
+  margin-top: 20px;
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
```
savagematt, line 66: Good to see you using css convenience methods like `repeat`

src/components/Footer.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+
+function Footer() {
+  const HotelContact = [
```
savagematt, line 4: By convention, `const` names start with lower case letters in javascript, so this would be `hotelContact`.

React function components are an exception to this general rule.

src/components/Heading.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import logo from "../Hotel Images/hotel.jpg";
+
+function Heading() {
+  return (
+    <header className="App-header">
+      <img className="logo" src={logo} alt="" />
```
savagematt, line 7: Remember to complete `alt` text so the site is accessible to blind users

src/components/SearchResults.js
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import Moment from "react-moment";
+function SearchResults({ results }) {
+  return (
+    <div>
+      <table className="table">
+        <thead>
+          <tr>
+            <th scope="col">Id</th>
```
savagematt, line 9: Great to see `scope` here for users of screen readers

savagematt (commented):
Great stuff Miguel.

Miguel-Cabral:
@savagematt  lesson 2 is completed

src/components/SearchResults.js
```diff
@@ -11,14 +11,22 @@ function SearchResults({ results }) {
...
+            const [select, setSelect] = useState(true);
+            function handleClick() {
+              setSelect(select => !select);
+            }
             return (
-              <tr key={index}>
+              <tr
+                key={index}
+                onClick={handleClick}
+                style={{ backgroundColor: select ? "" : "green" }}
```
savagematt, line 28: You can use `style` directly for something simple like this.

As the site gets larger and more complex, you'll probably have other tables which need rows to be highlighted. In that case you'll want to define a `selectedRow` css class once and then reuse it everywhere you need to highlight table rows (using `className` in React or just `class` if you're writing plain html). Then you can change the style of all tables at once, instead of the `style` property of every table in the site.

If you run the app, the grey text is quite hard to read against the dark green background, which isn't the best for usability. Maybe change the text colour to white, or make the green colour much lighter so the text still pops?

src/components/SearchResults.js
```diff
@@ -11,14 +11,22 @@ function SearchResults({ results }) {
...
-            <th scope="col">check out date</th>
+            <th scope="col">Check In Date</th>
+            <th scope="col">Check Out Date</th>
           </tr>
         </thead>
         <tbody>
           {results.map((person, index) => {
+            const [select, setSelect] = useState(true);
+            function handleClick() {
+              setSelect(select => !select);
```
savagematt, line 22: Using `!select` rather than `if(select) {return false;} else {return true;}`, for example, is good! Terse, but also readable.

src/components/SearchResults.js
savagematt: Ah, in fact bootstrap already has some classes you can use for this.

https://getbootstrap.com/docs/4.0/content/tables/#contextual-classes

PR #498, opened 2022-12-06
nw5-Manchester-Davood-Moradi-React-Week1

clav00cl:


src/Search.js
```diff
@@ -17,7 +18,8 @@ const Search = () => {
                 className="form-control"
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
+              {/* <button className="btn btn-primary">Search</button> */}
+              <SearchButton />
```
Dominic-Taylor-Dev, line 22: I don't necessarily agree with your choice to make `SearchButton` a child of `Search`. Let's discuss this when we speak. However, if you want to take this approach, I suggest something like the following:

1) Keep the 'knowledge' of the complete booking list `Bookings` component. You already keep a state of bookings via `const [bookings, setBookigs] = useState([]);` so you can keep that
2) Start keeping track of the search term, also at that higher level i.e. inside the `Bookings` component: `const [searchField, setSearchField] = useState("");`
3) Do the filtering action at the level of the `Booking` component with a variable like this (the component will re-render every time the state changes so you don't need to worry that this will get out of date):
```
  const filteredBookings = bookings.filter((booking) => {
    return (
      booking.firstName.toLowerCase().includes(searchField.toLowerCase()) ||
      booking.surname.toLowerCase().includes(searchField.toLowerCase())
    );
  });
```
4) Pass that `filteredBookings` rather than all bookings to your `SearchResult` (this will also re-render when you search, because `filteredBookings` will have a new value when the parent re-renders):
```
        <SearchResult
          filteredBookings={filteredBookings}
          showProfile={showProfile}
        />
```

5) Pass `setSearchField` (the new setState() we defined in point 2) down to the child components of `Bookings` - in this case `Search`. From here, you can define a function that gets the value of the input box whenever the function is called, and then sets the search field (all the way up in `Bookings`) with that value:
```
  const searchButtonSubmit = (event) => {
    event.preventDefault();
    const searchBarValue = document.getElementById("customerName").value;
    setSearchField(searchBarValue);
  };
```

6) Pass the new  `searchButtonSubmit` function as a prop to your `SearchButton`:
```
<SearchButton searchButtonSubmit={searchButtonSubmit} />
```
7) Add that prop as your click handler inside the Search Button class:
```
const SearchButton = ({ searchButtonSubmit }) => {
  return (
    <div>
      <button className="btn btn-primary" onClick={searchButtonSubmit}>
        Search
      </button>
    </div>
  );
};
```

Now, hopefully you can see that's quite a complicated way of doing things. I think there are simpler ways if we re-arrange our parents and children, but it's beyond the scope of a written code review for me to discuss that fully, so let's talk about it soon

Dominic-Taylor-Dev (commented):
It's a good start, well done. Things I like are the fact you've got the nights calculating well, or that you're populating data from an API. 

The thing I'd focus on figuring out next to improve this is how to make the search work properly in terms of being able to change the search term after you've clicked 'Search' for the first time. I've given you an implementation that stays close to your existing code, but there are better ways to do it in my opinion. The rest is okay, but you might also want to consider your layout/styling to make things align nicely

src/Bookings.js
```diff
@@ -1,21 +1,51 @@
...
+  const [customerProfile, setCustomerProfile] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
   const search = searchVal => {
-    console.info("TO DO!", searchVal);
+    const searchValue = searchVal.toLowerCase();
+    setBookigs(
+      bookings.filter(element => {
+        if (
+          searchValue === "" ||
+          element.firstName.toLowerCase().includes(searchValue) ||
```
Dominic-Taylor-Dev, line 16: Don't forget that the search value should also be `.toLowerCase()`. Otherwise it will only match when the user types in the search box only with lower case

PR #500, opened 2022-12-07
NW5 - Manchester - Yousef Shaikhan - react - week1

youweb3:


src/Lesson1/Order.js
```diff
@@ -0,0 +1,17 @@
...
+    setOrder(order + 1);
+  }
+  return (
+    <>
+      <li>
+        {OrderType}: {order}
+        <RestaurantButton orderOneProps={orderOne} />
+      </li>
+    </>
+  );
```
SteveLeicester, line 16: Good - the button is a separate component - but in effect you give the button something to do when it is clicked.

src/Lesson1/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export default function RestaurantButton({ orderOneProps }) {
+  return (
```
SteveLeicester, line 4: Nice destructuring of the props object being passed to the button.

src/Lesson1/SearchResults.js
```diff
@@ -0,0 +1,56 @@
...
+            const [color, setColor] = useState(false);
+            function changeColor() {
+              setColor(!color);
+            }
+            return (
+              <tr
+                key={index}
+                style={{ backgroundColor: color ? "yellow" : "white" }}
+                onClick={changeColor}
+              >
```
SteveLeicester, line 39: This is nice and simple - works well.  You can also add and remove a CSS class name to do the same. Often preferable since you can load a different style sheet if you decide on different colours. So we make class - note the helpful name.

```
.searchResultSelected{
  background-color: #FF0000;
}
```
So now in the code I no longer mention anything about colour but instead think in terms of selected or not selected.  I change the function to toggleSelected.  Search toggle switch in google - it usually means on or off but if it is on you can only turn off and if it is off you can only turn on.  Here we mean selected or not selected.

```
const [isSelected, setSelected] = useState(false);
function toggleSelected() {
     setSelected(!isSelected);
}
```
Now we can set the class name like so

```
<tr
      key={index}
      className={isSelected?"searchResultSelected":""}
      onClick={toggleSelected}
>

```

src/Restaurant.js
```diff
@@ -1,14 +1,18 @@
...
   return (
     <div>
       <h3>Restaurant Orders</h3>
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order key={1} OrderType={"pizzas"} />
+        <Order key={2} OrderType={"salads"} />
+        <Order key={3} OrderType={"chocolate cake"} />
```
SteveLeicester, line 15: Good we spoke about having unique keys.  I'm glad the restaurant has pizza too :-)

src/index.css
```diff
@@ -24,3 +24,19 @@ body {
...
 }
+
+.footer li {
+  list-style: none;
+}
+
+.card {
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
+}
+
```
SteveLeicester, line 35: The style of the cards looks really nice - like it.  Its very satisfying to create a very pleasing look.

SteveLeicester:
This is great stuff Yousef. The hotel project is coming along really nicely.  Thankyou for your hard work.

PR #522, opened 2023-02-23
London 9 - Lovelace - Shayan Mahnam - React - Week 1

ShayanMahnam:
### Demo
<a href="https://cyf-shayanmahnam-hotel-react.netlify.app/">Shayan Hotel React</a>


src/Order.js
```diff
@@ -0,0 +1,20 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = props => {
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    setOrders(preOrder => preOrder + 1);
```
sampennington, line 8: Nice!
I dont know if we covered this, could you explain why we do this? 😄

ShayanMahnam: Because it’s a way that is safe and guaranteed to be synchronous, this is important because updating the state in a synchronous way ensures that the updated value is immediately available for other components that depend on it, which is not guaranteed when you update a variable directly.

you covered this in week 2 before exercise D 😄

howard-ss: It's mentioned by Mr. Moe Ibrahim on his session on 03 Feb when we discussed the different options for state function expression, but it's not covered in the syllabus.

src/App.js
```diff
@@ -1,13 +1,25 @@
...
+import Restaurant from "./Restaurant";
+import Footer from "./Footer";
 import "./App.css";
 
 const App = () => {
+  const contactDetails = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
```
sampennington, line 14: It would be better to define this variable outside of the component, it is just a constant and doesn't need to be recreated every render 😄

src/Bookings.js
```diff
@@ -1,18 +1,77 @@
...
+          return res.json();
+        }
+      })
+      .then(data => {
+        setBookings(data);
+        setData(data);
+        setIsLoading(false);
+      })
+      .catch(error => {
+        console.log(error);
```
sampennington, line 38: Anything else that we could do in our catch function, would the user be told if the request failed like this?

src/Bookings.js
```diff
@@ -1,18 +1,77 @@
...
+      });
+  }, []);
+
   const search = searchVal => {
-    console.info("TO DO!", searchVal);
+    if (searchVal.trim() === "") {
+      setBookings(data);
+    } else {
+      const filterBookings = bookings.filter(element => {
```
sampennington, line 46: Instead of using `element` here, it might be clearer to use `booking`. This value is a `booking` from our `bookings` array! It makes it a bit more readable knowing what the value is 😄 .

src/Bookings.js
```diff
@@ -1,18 +1,77 @@
...
         <Search search={search} />
-        {/* <SearchResults results={FakeBookings} /> */}
+        {errorOccurred.ok === false ? (
+          <div>
+            <span>
+              An Error occurred when fetching the Bookings Data :{" "}
+              {errorOccurred.status} {errorOccurred.statusText}
+            </span>
+          </div>
+        ) : isLoading === true ? (
```
sampennington, line 67: Do we need to check if this `isLoading` boolean is equal to true here?  It only has two possible values

sampennington: Same with `ok` above 😄

src/CustomerProfile.js
```diff
@@ -0,0 +1,46 @@
+import React, { useState, useEffect } from "react";
+
+const CustomerProfile = props => {
+  const [customerProfileData, setCustomerProfileData] = useState("");
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${props.id}`)
+      .then(res => res.json())
+      .then(data => setCustomerProfileData(data));
```
sampennington, line 9: Could this request fail also? Always a good idea to handle errors 👍

src/Footer.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+const Footer = props => {
+  return (
+    <footer>
+      <ul className="footer-contact">
+        {props.contactDetails.map((element, index) => (
+          <li key={index}>{element}</li>
```
sampennington, line 8: It's best to try and avoid using the array `index` as a `key`. I know this is mentioned in the syllabus but have a read of this if you are not aware already https://legacy.reactjs.org/docs/lists-and-keys.html#keys

The react docs can do a much better explanation then me 😊

src/SearchResult.js
```diff
@@ -0,0 +1,48 @@
...
+const SearchResult = props => {
+  const [customClass, setCustomClass] = useState("");
+  const [showCustomerProfile, setShowCustomerProfile] = useState(false);
+
+  const classToggle = () => {
+    setCustomClass(customClass => (customClass ? "" : "table-primary"));
+  };
+
+  const toggleCustomerProfile = event => {
+    event.stopPropagation();
```
sampennington, line 13: What did we need to stop propagating here?

src/SearchResults.js
```diff
@@ -0,0 +1,48 @@
...
+            <th scope="col">Check in Date</th>
+            <th scope="col">Check out Date</th>
+            <th scope="col">Total Nights</th>
+            <th scope="col">Profiles</th>
+          </tr>
+        </thead>
+        <tbody>
+          {props.results.map(user => {
+            let a = moment(user.checkInDate, "YYYY-MM-DD");
+            let b = moment(user.checkOutDate, "YYYY-MM-DD");
```
sampennington, line 25: Think we can use `const` for these? Generally a better idea to use `const` if we don't plan on reassigning something.

sampennington (commented):
Amazing work Shayan! 🚀 
Just a few little points ☺️

PR #524, opened 2023-02-23
London 9 - Natalie Zablotska - React - Week 1

nataliiazab:


nataliiazab:
Hi @Hamisakim! If you have some extra time - please write your suggestions to improve the code, if any. Thanks!!

Hamisakim:
hi, @nataliiazab are there any specific parts or bugs, or areas you were troubled with? I'll have a look :)

src/App.css
```diff
@@ -52,6 +55,56 @@ tr {
...
+  .card {
+    flex-basis: calc(50% - 10px);
+  }
+}
+
+@media screen and (max-width: 480px) {
+  .card {
+    flex-basis: 100%;
+  }
 }
```
sampennington, line 80: This is great, thinking about responsivness i love it ❤️

src/App.css
```diff
@@ -52,6 +55,56 @@ tr {
...
+.selectRow {
+  background: red;
+}
+
+.restaurant {
+  text-align: center;
+  margin: 4rem 15rem;
+ 
+}
+.restaurant-orders {
```
sampennington, line 103: Try and leave a line between these classes, it helps it be a bit neater.

src/App.js
```diff
@@ -2,14 +2,29 @@ import React from "react";
...
+import Footer from "./Footer";
+import Restaurant from "./Restaurant";
+
 
 const App = () => {
+  const contactDetails = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789",
+  ];
```
sampennington, line 16: It is great we have put this into a variable, however it could be better to define this outside of the component itself. Right now we are creating a new variable every time it renders, but it never changes! 😄

src/Bookings.jsx
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return response.json();
+      })
+      .then((data) => {
+        setBookings(data);
+        setIsLoading(true);
+      })
+      .catch((error) => {
+        console.log(error + "in file Bookings.jsx");
+        setError(error);
```
sampennington, line 71: We have considered both types of errors our `fetch` function can result in. Good work 😎

src/Bookings.jsx
```diff
@@ -0,0 +1,69 @@
...
+    fetch(`https://cyf-react.glitch.me/delayed`)
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error("HTTP error, status = " + response.status);
+        }
+        return response.json();
+      })
+      .then((data) => {
+        setBookings(data);
+        setIsLoading(true);
```
sampennington: Does this work how it should? I think we should be setting loading state to false after our api call has finished 🤔

src/Bookings.jsx
```diff
@@ -0,0 +1,69 @@
...
+        ) : (
+          <p></p>
+        )}
+        {error !== null ? (
+          <p>
+            We are sorry, it is not possible to download data right now. Please
+            try again later{" "}
+          </p>
+        ) : (
+          <p></p>
```
sampennington: We don't want to return an empty paragraph element when there is no error, what else could we do here?

I think one way would be to not use a ternary operator at all here, we could just do `error && `

src/Bookings.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }, []);
+
+  return (
+    <div className="App-content">
+      <div className="container">
+        <Search setSearchInput={setSearchInput} searchInput={searchInput} />
+        {isLoading === false && error === null ? (
+          <p>Data is loading...</p>
+        ) : (
+          <p></p>
```
sampennington: Same as below, we dont want an empty paragraph element here!

src/Bookings.jsx
```diff
@@ -0,0 +1,69 @@
...
+        console.log(error + "in file Bookings.jsx");
+        setError(error);
+      });
+  }, []);
+
+  return (
+    <div className="App-content">
+      <div className="container">
+        <Search setSearchInput={setSearchInput} searchInput={searchInput} />
+        {isLoading === false && error === null ? (
```
sampennington: `isLoading` is a boolean, it can only have two values, true and false. An easier way of writing this would be:

```
!isLoading && !error && <p>Data is loading...</p>
```

sampennington: I just noticed, this loading state is the wrong way round, when loading is true it should show the loading text, I think this is because you do:
`  setIsLoading(true);` after the api call like i said above!

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from "react";
+
+const CustomerProfile = (props) => {
+  const [customer, setCustomer] = useState("");
+
+  let id = props.id;
```
sampennington, line 6: Think we can use `const` here, always best to use const when we know our variable is not going to reassigned.

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
...
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then((response) => response.json())
+      .then((data) => setCustomer(data))
+      .catch((err) => console.log(err));
+  }, [id]);
+  if (customer.id) {
```
sampennington, line 14: Best to try and leave a space between these, so its easier to seperate whats going on 😄

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
...
+  const [customer, setCustomer] = useState("");
+
+  let id = props.id;
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then((response) => response.json())
+      .then((data) => setCustomer(data))
+      .catch((err) => console.log(err));
+  }, [id]);
```
sampennington, line 13: Nice including `id` in the dependency array 🥳

src/CustomerProfile.js
```diff
@@ -0,0 +1,30 @@
...
+        <p>Id: {customer.id}</p>
+        <p>Email: {customer.email}</p>
+        {customer.vip && <p>Phone number: {customer.phoneNumber}</p>}
+      </div>
+    );
+  } else {
+    return "";
+  }
+};
+
```
sampennington, line 29: A simpler way or structuring this component would be having:

```
if(!customerId) {
  return null
}
```
Above the main return. Then you dont have to have if/else.

It is also more common practice to return `null` from a component that we don't want to render anything, than an empty string 😄

src/App.js
```diff
@@ -2,14 +2,29 @@ import React from "react";
...
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCard />
       <Bookings />
+      <Restaurant />
+      <Footer props={contactDetails} />
     </div>
   );
 };
```
sampennington, line 27: This is a great structure, so easy to see how everything fits together! 😊

src/App.js
```diff
@@ -2,14 +2,29 @@ import React from "react";
...
+    "0123 456789",
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCard />
       <Bookings />
+      <Restaurant />
+      <Footer props={contactDetails} />
```
sampennington, line 23: Its best not to call a prop, `props`. What you could do here is `{...contactDetails}` this would pass all of the contact details down to the Footer as seperate props. Otherwise in your component you have `props.props.xxx` which loooks a bit rubbish! 😅

src/TouristInfoCards.jsx
```diff
@@ -0,0 +1,74 @@
...
+          institutions including the Scottish Opera, Scottish Ballet and
+          National Theatre of Scotland, as well as acclaimed museums and a
+          thriving music scene.
+        </p>
+        <div className="card-btn-container">
+          <a href="http://peoplemakeglasgow.com/" target="_blank">
+            <button className="card-btn">More information</button>
+          </a>
+        </div>
+      </div>
```
sampennington, line 27: There is a lot of duplication in this file, what can we do when we see duplicated html with different values?

sampennington (commented):
Awesome work Natalie!

Really enjoying seeing how much progress you are making ❤️  it is really great to see 😄 

If you have any questions on anything I have said, do reach out to me, or the help center.

PR #525, opened 2023-02-24
Inna-Poliakova_London_9_cyf-hotel-react 

Innapoliakova:


src/Order.js
```diff
@@ -0,0 +1,21 @@
+import React, {useState} from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = ({ orderType }) => {
```
sampennington, line 4: I like the use of restructuring here Inna 🥳

src/SearchResults.js
```diff
@@ -0,0 +1,66 @@
...
+  const [selectedRows, setSelectedRows] = useState([]);
+  // create a new state variable 'selectedRows' using the useState hook.
+//  here will be store the selected rows 
+// we start from empty array
+
+  const handleRowClick = (id) => {
+    if (selectedRows.includes(id)) {
+      setSelectedRows(selectedRows.filter(rowId => rowId !== id));
+    } else {
+      setSelectedRows([...selectedRows, id]);
```
sampennington: This is a cool way of doing this. 
Can you think of another way where we just keep a `selected` state, which is equal to true or false?

PR #527, opened 2023-02-24
Mo

momahboobian:


src/Nights.jsx
```diff
@@ -0,0 +1,8 @@
+import moment from "moment";
+
+export const numberOfNight = ({ checkInDate, checkOutDate }) => {
+  const checkIn = moment(checkInDate);
+  const checkOut = moment(checkOutDate);
+  const nights = checkOut.diff(checkIn, "days");
+  return nights;
+};
```
sampennington, line 8: This is great, breaking logic from components out into separate functions 🥳

src/SearchResults.jsx
```diff
@@ -0,0 +1,45 @@
...
+const SearchResults = ({ results }) => {
+  const [selectedRow, setSelectedRow] = useState(null);
+
+  const handleRowClick = (id) => {
+    if (id === selectedRow) {
+      setSelectedRow(null);
+    } else {
+      setSelectedRow(id);
+    }
+  };
```
sampennington, line 13: I dont think this will work for this part will it?
```
Test: Verify that each row of your table can be highlighted (on and off) independently when being clicked.
```
How might we make the rows be highlighted independently ?

sampennington (commented):
Looks great Mo 🥳

src/SearchResults.jsx
momahboobian: Sorry @sampennington,
I did not understand, at the moment it work. But if it's necessary or not! is that the question?

PR #528, opened 2023-02-25
London9_Lovelace_Meysam_Arshadi_React_Week1_CYF_Hotel_React

Meysam-Arshadi:


src/Footer.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+let footerProp = [
+  "123 Fake Street, London, E1 4UD",
+  "hello@fakehotel.com",
+  "0123 456789",
+];
```
sampennington, line 6: This doesnt appear to be being used as a prop, did you mean to pass it as a prop from App?

If not, best not to end a variable in the word `prop`, as thats a well known word for props in react 😄

src/TouristInfoCards.js
```diff
@@ -0,0 +1,57 @@
...
+            href="https://peoplemakeglasgow.com"
+            className="btn btn-primary"
+            target="_blank"
+          >
+            Go somewhere
+          </a>
+        </div>
+      </div>
+    </div>
+  );
```
sampennington, line 54: Not part of the homework, but theres an improvement that could be made to this file to avoid duplication, can you see it?

sampennington (commented):
Looks great Meysam could only find a couple comments to leave 😄

src/App.css
```diff
@@ -1,42 +1,40 @@
...
-  height: 80px;
+  font-family: "Segoe UI", Tahoma, Geneva, Verdana, sans-serif;
 }
 
 .App-header {
-  background-color: #222;
-  height: 50px;
-  padding: 20px;
+  display: flex;
+  flex-direction: row;
```
sampennington: The default flex direction is row, you shouldn't need to add this.

src/Footer.js
Meysam-Arshadi: Thanks for pointing that out, I haven't noticed that. :)
The file was created on the original template and I think it will be used later on at the stretch exercises.

src/App.css
Meysam-Arshadi: Thanks! Corrected.

src/TouristInfoCards.js
Meysam-Arshadi: Thanks Sam. I am not sure, but should I create an array of objects from the src(s) and pass them to a new function called creatCard( ) ?

sampennington: I was jsut meaning you could create a new component called InfoCard and pass some props to it :)

PR #531, opened 2023-02-25
London Class 9 - Farzaneh Haghani - React - Week 1

farzaneh-haghani:
https://cyf-farzaneh-haghani-hotel-react.netlify.app


src/App.js
```diff
@@ -1,13 +1,24 @@
...
+      <Heading />
+      <TouristInfoCards cities={fakeCities} />
       <Bookings />
+      <Footer
+        arr={[
+          "123 Fake Street, London, E1 4UD",
+          "hello@fakehotel.com",
+          "0123 456789"
+        ]}
+      />
```
JDysiewicz: I'd pass each of this as a separate prop rather than as an array together - e.g.
```
<Footer address-{'123 FakeStreet'} email={'hello@fakemail'} phoneNumber={'0123 45678' />}
```
just because they're all strings than represent something specific rather than all the same thing

src/App.js
```diff
@@ -1,13 +1,24 @@
...
+import Footer from "./Footer";
 
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards cities={fakeCities} />
       <Bookings />
+      <Footer
```
JDysiewicz: Beautiful structure, good job!

src/Footer.js
```diff
@@ -0,0 +1,9 @@
+const Footer = props => (
+  <ul className="footer">
+    {props.arr.map((eachItem, index) => (
+      <li key={index}>{eachItem}</li>
```
JDysiewicz: Good job remembering the `key` property, many even professionals forget this! A note of caution though, try not to use the `index` of the array as the key, because (especially when you start dealing with state in arrays) the entry found at each index can change, but because the index doesn't change React won't re-render the entry in the array. The `key` should uniquely identify each element in an array.

src/Heading.js
```diff
@@ -0,0 +1,11 @@
+const Heading = () => (
+  <header className="App-header">
+    <img
+      className="App-logo"
+      src="https://img.freepik.com/premium-vector/hotel-logo-simple-illustration_434503-736.jpg?w=2000"
+    />
+    <span className="App-title">CYF Hotel</span>
+  </header>
```
JDysiewicz: Perfect, small improvement could be to add the `imageSrc` and `spanText` as props - but it's also fine to leave as-is if they're never going to be reused.

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
...
+import moment from "moment";
+
+const SearchResults = props => (
+  <table className="table table-striped">
+    <CreateTableHeader />
+    <CreateEachRow results={FakeBookings} />
+  </table>
+);
+
+const CreateTableHeader = () => (
```
JDysiewicz: Really really nice idea here, I like it. Small adjustment would be to name this `TableHeader` as it literally _is_ the tableheader rather than `CreateTableHeader`. Another small adjustment would be to create a `Table` component which uses the `TableHeader` and `TableRow` compeonnts. e.g.

```
const SearchResults = () => {
   return <Table />
}

const Table = () => {
   return (
      <table className=''>
         <TableHeader />
         <TableRow data={fakeBookings}
      </table>
   )
}
```

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
+import FakeBookings from "./data/fakeBookings.json";
```
JDysiewicz: `PascalCase` is reserved for React components, you're importing a JSON file so this should be `camelCase` as it's a normal javascript variable

src/data/fakeCities.json
```diff
@@ -0,0 +1,20 @@
...
+    "description": "Manchester is the nucleus of the largest metropolitan area in the north of England, and it remains an important regional city, but it has lost the extraordinary vitality and unique influence that put it at the forefront of the Industrial Revolution.",
+    "information": "https://visitmanchester.com"
+  },
+  {
+    "name": "London",
+    "image": "https://www.fodors.com/assets/destinations/2869/tower-bridge-london-england_980x650.jpg",
+    "description": "London is among the best known writings by visionary English poet William Blake. The poem describes a walk through London, which is presented as a pained, oppressive, and impoverished city in which all the speaker can find is misery.",
+    "information": "https://visitlondon.com"
+  }
+]
```
JDysiewicz, line 20: Beautiful use of splitting your data out into a separate file - this will make it much easier to adjust your existing code to work with hitting an API to get this data instead, or getting it from some other file etc. Really nice touch that shows you're thinking about how to make your code more maintainable in the long-run

JDysiewicz (commented):
A really really solid piece of work! You can tell you spent a lot of time styling this, it looks great (good job on deploying to netlify!). The structure of your code makes sense and is easy to read. There are a few small things I'd adjust, but what you've produced is really solid well done!

src/data/fakeCities.json
Bahare09: perfect, well done Farzane

src/Bookings.js
```diff
@@ -1,18 +1,36 @@
...
+        // console.log(data);
+      })
+      .catch(err => console.log(err));
+  }, []);
+
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    setBooking(
+      booking.filter(
+        eachPerson =>
```
sampennington, line 23: I think we could use a better variable name here, `eachPerson` doesn't really reflect what this variable is. It is a `booking`, from our `bookings` array, so maybe we could try calling it `booking` 😄

src/CustomerProfile.jsx
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from "react";
+
+const CustomerProfile = ({ id } = props) => {
+  const [data, setData] = useState("");
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then(res => res.json())
+      .then(data => setData(data))
+      .catch(err => console.log(err));
```
sampennington, line 10: Love the error handling 😄

src/RestaurantButton.jsx
```diff
@@ -0,0 +1,9 @@
+const RestaurantButton = props => {
+  return (
+    <button onClick={props.onClick} className="btn btn-primary">
+      Add
+    </button>
+  );
+};
+
```
sampennington, line 8: Love the small components, so easy to see what is going on! 😄

src/TouristInfoCards.jsx
```diff
@@ -0,0 +1,18 @@
+const TouristInfoCards = props => (
```
sampennington, line 1: This only returns one `Card`, it is good to be as clear as possible what a component is going to return, so I would suggest losing the S on the end of it.

sampennington (commented):
Great work Farzaneh!!

PR #533, opened 2023-02-26
London 9 - Oleh Pysmenko - React - Week 1

tiritaki:
https://cyf-tiritaki-hotel-react.netlify.app/


src/Footer.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+function Footer (props){
+        const li = props.address.map((element) => <li> {element} </li>)
+        return (
+        <ul className="unorderList"> {li}</ul>
+    )
```
sampennington, line 7: I love the small simple components, so easy to see whats gone on! 😎

src/App.js
```diff
@@ -1,13 +1,19 @@
 import React from "react";
+import Heading from "./Heading";
+import TouristInfoCards from "./TouristInfoCards";
+import Footer from "./Footer"
 
 import Bookings from "./Bookings";
 import "./App.css";
 
 const App = () => {
+  const addressArray = ["123 Fake Street, London, E1 4UD", "hello@fakehotel.com", "0123 456789"];
```
sampennington: When calling variables I would tend to leave the type our of the name of the variable, I thin `addresses` is a good name here 😄

src/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
...
+          className="card-img-top"
+        />
+        <h1>London</h1>
+        <p>This is a description</p>
+        <div className="card-body">
+          <a href="visitlondon.com" className="btn btn-primary">
+            More Information
+          </a>
+        </div>
+      </div>
```
sampennington, line 21: We have a lot of duplication in this file, I wonder if we know a way to reuse some of it?
😄

src/TouristInfoCards.js
```diff
@@ -0,0 +1,49 @@
...
+    <div className="cards-wrapper">
+      <div className="card">
+        <img
+          src="https://media.timeout.com/images/105400989/750/422/image.jpg"
+          className="card-img-top"
+        />
+        <h1>Glasgow</h1>
+        <p>This is a description</p>
+        <div className="card-body">
+          <a href="peoplemakeglasgow.com" className="btn btn-primary">
```
sampennington: You are missing something on this link, it wont work, can you see it?

sampennington (commented):
Awesome work Oleh! Few comments but nothing major 🥳

src/Footer.js
tiritaki: thank you

src/App.js
tiritaki: thank you for advice. I just want to avoid writing <Footer address = {address}/>

src/TouristInfoCards.js
tiritaki: you are right, I could create an array of objects with card info and  dynamically create each card as I did for footer and table but there were no such requirements in the task, so I omitted it, maybe in future will change it

src/TouristInfoCards.js
tiritaki: fixed it, added https:// and target="_blank"

src/TouristInfoCards.js
sampennington: I was meaning we could made a `TouristInfoCard` component which we could use three times

sampennington: Yes not aware of any requirement to do so, just a tip! 😄

tiritaki: Cool, thank you :)

src/SearchResults.js
```diff
@@ -0,0 +1,41 @@
+import moment from 'moment';
+import SearchRow from './SearchRow';
+
+function SearchResults(props) {
+  
+  let tr = props.results.map((element) => (
+    <SearchRow
+       key = {element.id}
+       id = {element.id}
```
sampennington: I haven't seen props passed quite like this before, it'd generally better ot just do
`id={element.id}` over `id = {element.id}`

src/SearchResults.js
```diff
@@ -0,0 +1,41 @@
+import moment from 'moment';
+import SearchRow from './SearchRow';
+
+function SearchResults(props) {
+  
+  let tr = props.results.map((element) => (
+    <SearchRow
+       key = {element.id}
```
sampennington: Nice, you have remembered to use a key! 😄

src/SearchResults.js
```diff
@@ -0,0 +1,41 @@
+import moment from 'moment';
+import SearchRow from './SearchRow';
+
+function SearchResults(props) {
+  
+  let tr = props.results.map((element) => (
```
sampennington: We are using `let` here, can we think of something better 🤔

src/SearchRow.js
```diff
@@ -0,0 +1,25 @@
+import React, {useState} from "react";
+
+function SearchRow (props){
+    const [selected, SetColour] = useState("")
```
sampennington: This will work great!  But it is generally accepted these variables follow a naming convention, `variable` and `setVariable`.

- There a couple improvements to be made here, first decide on what this state is, is it `selected`, or is it a `colour`?
- Then the `SetColour` need to start with a lower case `S`, we should name variables using camel case. 😄

sampennington (approved):
Nice one Oleh, you seem to have grasped all of this really well 😊 

My only comments are around naming conventions and styling really, well done!

src/SearchRow.js
tiritaki: also changed it, thank you for advice

src/SearchResults.js
tiritaki: using const now

src/SearchResults.js
tiritaki: I do remember

src/SearchResults.js
tiritaki: yes, I fixed it

PR #536, opened 2023-02-27
London9-Lovelace-Mohamed Abdi-cyf-hotel-react

MohamedAbdi114:
@JDysiewicz

### DIFF
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -47,6 +47,7 @@ const Bookings = () => {
       .catch(err => {
         console.log(err.message);
         setError(true);
+        setDataLoading(false)
       });
   }, []);
 
@@ -68,7 +69,9 @@ const Bookings = () => {
       <div className="container">
         <Search search={search} />
         {/* <SearchResults results={FakeBookings} /> */}
-        <SearchResults results={bookings} />
+        <div>{dataLoading&&<span>Loading..</span>}</div>
+        <div>{error&&<span>there is an Error..</span>}</div>
+        <SearchResults  results={bookings} />
       
         
 
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, {useState} from "react";
 import Footer from "./Footer.js";
 import Search from "./Search.js";
 import SearchResults from "./SearchResults.js";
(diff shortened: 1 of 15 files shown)


src/Bookings.js
```diff
@@ -13,6 +22,19 @@ const Bookings = () => {
       <div className="container">
         <Search search={search} />
         {/* <SearchResults results={FakeBookings} /> */}
+        <SearchResults results={FakeBookings} />
+       
+        
+        <TouristInfoCards  town="Visit Glasgow"  link = "http://peoplemakeglasgow.com" imageSource = "https://peoplemakeglasgow.com/imager/general/109009/City-centre-George-Square-Content-Block_aa93d8ee9a3cc573b9a01b328410e8b2.jpg"/>
+        <TouristInfoCards town = "Visit Manchester" link = "http://visitmanchester.com" imageSource = "https://media.wired.co.uk/photos/6357b98e09e6f1942a2a8a9a/16:9/w_1280,c_limit/HSBC%20MANCHESTER.jpeg"/>
+        <TouristInfoCards town = "Visit London" link = "http://visitlondon.com" imageSource = "https://cdn.londonandpartners.com/-/media/images/london/visit/campaigns/international-recovery-campaign/lets-do-london-related-international-640.jpeg?mw=640&hash=477E6FF9F31F292EE017E5BA63027C7054EB39A0"/>
```
JDysiewicz: Really nice use of a component here to reduce HTML repetitiveness. One small change you could make is to make this information an array of objects, and then `.map` over that array to render out the results. This means if you extend this to get this information from an API, or want to add a new town, you just need to add a row to that array rather than manually write more JSX. e.g.
```
const townArr = [
   {titleText: 'Visit Glasgow', link: 'http://...', imgLink: 'http://...'},
   {titleText: 'Visit Manchester', link: 'http://...', imgLink: 'http://...'},
]

const  Bookings  = () => {
   ...

   return (
      ...
      { townArr.map(({titleText, link, imgLink}) => <TouristInfoCard key={titleText} town={titleText} link={link} img={imgLink}>) }
   )
}

```

src/Footer.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+const Footer = () => {
+  let address = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
```
JDysiewicz, line 8: Could probably use `const` here instead of `let`

src/Footer.js
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <div>
+      <ul>
+        {address.map(item => {
+          return <li>{item}</li>;
```
JDysiewicz: When doing `.map` to return JSX, you need to add a `key` property to the parent element as this is needed for React to know when it should update if the array changes. e.g.
```
{address.map(item => return <li key={item}>{item}</li>)}
```

src/SearchResults.js
```diff
@@ -0,0 +1,43 @@
...
+              <td>{booking.id}</td>
+              <td>{booking.title}</td>
+              <td>{booking.firstName}</td>
+              <td>{booking.surname}</td>
+              <td>{booking.email}</td>
+              <td>{booking.roomId}</td>
+              <td>{booking.checkInDate}</td>
+              <td>{booking.checkOutDate}</td>
+              <td>
+                {moment(booking.checkOutDate).diff(booking.checkInDate, "days")}
```
JDysiewicz: Really good to see you making use of external libraries to solve problems, it's a good habit to get in to. A very small nitpick is that you can already do this in native JS with `checkoutDate.getDate() - checkInDate.getDate()`, as the javascript `Date` object allows things like subtracting/adding for dates.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,19 @@
...
+       <div className="card">
+       
+         <img src={props.imageSource} className="card-img-top" />
+            <div className="card-body">
+            <a href={props.link} className="btn btn-primary">{props.town}</a>
+           </div>
+           </div>
+           
+           );
+           }
```
JDysiewicz: perfect use of a component, good job. Are you using `prettier` to format this? looks like the indentation has gone wonky

JDysiewicz (commented):
Really good usage of components, showing a good understanding of the theory behind React. Only a few small things I'd change, which is certainly beyond the level you're expected of right now, so really good job!

MohamedAbdi114:
@JDysiewicz

MohamedAbdi114:
@JDysiewicz

src/Bookings.js
```diff
@@ -1,34 +1,83 @@
...
+        }
+        throw Error("Error fetching data");
+      })
+      .then(data => {
+        setBookings(data);
+        setDataLoading(false);
+      })
+      .catch(err => {
+        console.log(err.message);
+        setError(true);
```
JDysiewicz, line 49: You'll also want to set `setDataLoading(false` when there's an error too, as it's completed the request

src/CustomerProfile.js
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${props}`)
+      .then(response => {
+        return response.json();
+      })
+      .then(data => {
+        setProfile(data);
+        console.log(props);
+      });
+  }, []);
```
JDysiewicz, line 15: This is good, good job - as an improvement notice how this `useEffect` only exists to populate `profile` with data? You could extract both this `useState` and `useEffect` to a custom hook (this is far beyond what is expected)

src/Restaurant.js
```diff
@@ -1,15 +1,32 @@
...
       <h3>Restaurant Orders</h3>
-      <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <ul>
+       
+        <Order orderType={"Pizzas"} />
+        <Order orderType={"Salads"} />
+        <Order orderType={"Chocolate cake"} />
```
JDysiewicz, line 27: As a slight improvement, you could instead create an array of orders `const orders = ["Pizzas", "Salads", "Chocolate cake"]` and then `{ orders.map(order => <Order orderType={order} />) }` to render out various orders

src/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React, { useState } from "react";
+
+function  RestaurantButton (props) {
+  return (
+    <button onClick={props.orderOne} className="btn btn-primary">
```
JDysiewicz, line 5: nice use of passing in an onClick handler to a component - you'll use this pattern a lot

src/SearchResults.js
```diff
@@ -15,12 +25,23 @@ const SearchResults = ({ results }) => {
...
+              
           return (
-            <tr>
+          
+         <tr  >
+         
+          
               <td>{booking.id}</td>
               <td>{booking.title}</td>
               <td>{booking.firstName}</td>
```
JDysiewicz: Nice idea here, although you could go one step further and make everything in that `return` it's own `TableRow` component

JDysiewicz (commented):
All looking good - I'd recommend you try a hand at making a custom hook to encapsulate the "set some state, useEffect to update the state" logic to really push yourself

PR #539, opened 2023-02-27
React-Ebrahim Beiati-Asl cyf-hotel

ebrahimbeiati:


src/App.js
```diff
@@ -1,15 +1,26 @@
...
+      <Restaurant/>
+      <SearchResults />
+      <Order/>
+      <Footer />
     </div>
   );
 };
-
-export default App;
+export default App();
```
JDysiewicz: Careful with this - you _want_ to export the whole function, `App`, whereas here you're only exporting the _return_ of calling `App()`. When writing components, React will handle the calling of the function for you when you use the `<Component />` syntax.

src/Bookings.js
```diff
@@ -1,18 +1,18 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
```
JDysiewicz, line 5: Small stylistic thing, `PascalCase` is often reserved for React components, with normal JS variables being `camelCase` - you're importing a JSON file here so this should be `camelCase` as it is not a React component (it can look confusing as people will assume `FakeBookings` is a React component). This should be `import fakeBookings from "./data/fakeBookings.json"`

src/Footer.jsx
```diff
@@ -0,0 +1,22 @@
...
+    "hello@fakehotel.com", 
+    "0123 456789"
+];
+
+
+const Footer = () => {
+  return (
+    <footer>
+      <ul className="footer-contact">
+        {props.map((props) => (
```
JDysiewicz, line 13: The variable name `props` is reserved for the argument passed to all React components (e.g. `const Footer = (props) => {...}`. Try not to call other variables `props` except for that one specific case, as it can be confusing. In this case, if you want to have default arguments for your props, you could use this syntax:
```
const Footer = (props = {address: "123 Fake Street", email: "hello@fakehotel.com", phoneNumber: "0123 456789"}) => {...}
```

or

```
const defaultProps =  {address: "123 Fake Street", email: "hello@fakehotel.com", phoneNumber: "0123 456789"};

const Footer = (props = defaultProps) => {...}
```

src/Heading.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+const Heading = () => {
```
JDysiewicz, line 3: Perfect!

src/Order.jsx
```diff
@@ -0,0 +1,20 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = (props) => {
+  const [orders, setOrders] = useState(0);
+
+  function orderOne() {
+    setOrders(orders + 1);
+  }
```
JDysiewicz: Perfect usage of `useState`, good job!

src/RestaurantButton.jsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+const RestaurantButton = props => {
+    return (
+        <button onClick={props.orderOne} className="btn btn-primary">Add</button>
```
JDysiewicz: Small thing, de-structuring out your props is normally good practice e.g. 
```
const RestaurantButton = ({orderOne}) => {
   return (
      <button onClick={orderOne} className="">Add</button>
   )
}
```

src/SearchResults.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+
+const SearchResults = ({ bookings }) => {
```
JDysiewicz: Nice use of destructuring here

src/TouristInfoCards.jsx
```diff
@@ -0,0 +1,64 @@
...
+          <h5 className="card-title">Glasgow</h5>
+          <p className="card-text">
+            Glasgow's city centre is home to flagship stores, impressive
+            shopping centres and designer within an easily walkable area.
+          </p>
+          <a href="http://peoplemakeglasgow.com" className="btn btn-primary">
+            Visit Glasgow
+          </a>
+        </div>
+      </div>
```
JDysiewicz: It looks like this set of HTML is repeated a few times - this could be extracted into a separate component and then rendered multiple times rather than repeating all this HTML

JDysiewicz (commented):
Looking really good, a few stylistic things I'd change but overall a really nice first set of work for React!

src/Bookings.js
```diff
@@ -1,18 +1,54 @@
...
+    fetch("https://cyf-react.glitch.me")
+      .then((response) => response.json())
+      .then((data) => setBookings(data))
+      .catch((err) => {
+        console.log(err);
+      });
+  }, []);
+
+  const [bookings, setBookings] = useState([]);
+  const [showList, setShowList] = useState(true);
```
JDysiewicz: Usually you'd put all the `useState`s above the `useEffect`s, though that's just a stylistic thing

src/CustomerProfile.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useEffect } from "react";
+
+const CustomerProfile = (props) => {
+  const [customerData, setCustomerData] = useState("");
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${props.id}`)
+      .then((res) => res.json())
+      .then((data) => setCustomerData(data));
+  }, [props.id]);
```
JDysiewicz, line 10: This could definitely be a good use for a custom hook, have a go at making one! :) (Don't worry, this is beyond what is expected)

src/Nights.jsx
```diff
@@ -0,0 +1,8 @@
+import moment from "moment";
+
+export const numberOfNight = ({ checkInDate, checkOutDate }) => {
+  const checkIn = moment(checkInDate);
+  const checkOut = moment(checkOutDate);
+  const nights = checkOut.diff(checkIn, "days");
+  return nights;
+};
```
JDysiewicz, line 8: Because you're not returning any `jsx` (this isn't a react component), this should just be a regular `.js` file instead of a `.jsx` file

src/Restaurant.js
```diff
@@ -1,17 +1,15 @@
...
-          <Order orderType="Pizza" />
-          <Order orderType="Salads" />
-          <Order orderType="Chocolate cake" />
-        </ul>
-      
+      <ul className="order-type">
+        <Order orderType="Pizza" />
+        <Order orderType="Cesar-Salad" />
+        <Order orderType="Chocolate cake" />
+        <Order orderType="Ice-Cream"/>
```
JDysiewicz, line 11: You could instead create an array of order types `const orders = ["Pizza", "Cesar sald", "Ice cream"]` and `{orders.map(order => <Order orderType={order} />)}` to save some repetition

src/SearchResults.js
```diff
@@ -1,35 +1,79 @@
...
+
+const SearchResults = ({ results }) => {
+  const [selectedRow, setSelectedRow] = useState(null);
+  const [customerData, setCustomerData] = useState(null);
+
+  useEffect(() => {
+    if (selectedRow) {
+      fetch(`https://cyf-react.glitch.me/customers/${selectedRow}`)
+        .then((response) => response.json())
+        .then((data) => setCustomerData(data));
```
JDysiewicz, line 13: make sure to add a `.catch` as this network request could fail, and you'll want some way of letting the user know that the request has failed

src/SearchResults.js
```diff
@@ -1,35 +1,79 @@
...
+                    checkInDate: result.checkInDate,
+                    checkOutDate: result.checkOutDate,
+                  })}
+                </td>
+                <td>
+                  <button onClick={() => handleRowClick(result.id)}>
+                    Show profile
+                  </button>
+                </td>
+              </tr>
```
JDysiewicz: This could all be its own `TableRow` component

JDysiewicz (commented):
Some really nice work, good job :)

PR #540, opened 2023-02-27
London 9 - Vitalina Kuzmenko - cyf-hotel-react

VitalinaKuzmenko:
deployment on Netlify: https://checkinn-hotel-react-website.netlify.app/


src/NewBookingForm.js
```diff
@@ -0,0 +1,149 @@
...
+  const updateEmail = (event) => {
+    setEmail(event.target.value);
+  };
+  const updateTitle = (event) => {
+    setTitle(event.target.value);
+  };
+  const updateCheckInDate = (event) => {
+    setCheckInDate(event.target.value);
+  };
+  const updateCheckOutDate = (event) => {
```
sampennington: We have a lot of very similar looking functions here, could you think of anyway to generalise them and make it reusable for all of these fields?

src/App.js
```diff
@@ -1,13 +1,30 @@
...
+    "0123 456789",
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
       <Bookings />
+      <div className="info-cards">
+        {Cities.map((city) => {
+          return <TouristInfoCards key={city.id} city={city} />;
```
sampennington: I would say that this would make sense to be called `TouristInfoCard` over `TouristInfoCards` as it is just one card. If we did the map inside the card i would call it what you did ☺️

src/BookingsRow.js
```diff
@@ -0,0 +1,52 @@
...
+      );
+      const data = await response.json();
+      setFinalData(data);
+    };
+    fetchGuest();
+  }, [guest.id]);
+
+  const calculateNights = useCallback((firstDate, secondDate) => {
+    let first = moment(firstDate, "YYYY-MM-DD");
+    let second = moment(secondDate, "YYYY-MM-DD");
```
sampennington, line 21: It is generally a better idea to use `const` where we can, if we are not going to be reassigning variables.

package.json
```diff
@@ -14,6 +16,7 @@
     "eject": "react-scripts eject"
   },
   "devDependencies": {
+    "eslint-plugin-react-hooks": "^4.6.0",
```
sampennington, line 21: Nice! I don't think we covered this? But its a great idea to follow these linting rules!

package.json
```diff
@@ -14,6 +16,7 @@
...
     "husky": "^8.0.3",
     "prettier": "^2.8.4",
     "pretty-quick": "^3.1.3"
```
sampennington, line 24: Do we we have our dependencies int he correct sections? 😃

sampennington (commented):
This is crazy good Vitalina!! I'm so impressed 😮
 I love how much care you have put into it all, from the way you've written functions, to commit messages and taking error handling into account 🥳 

On thing I did notice, there are quite a few instances of using `let` over `const`, where it's not wrong, its a lot more common to use `const` for variables we are not reassigning, did you do this intentionally?

src/NewBookingForm.js
VitalinaKuzmenko: Thank you, Sam! I've already created one function that saves all input values inside the object and then pass the object to handleSubmit function.

src/App.js
VitalinaKuzmenko: Yes, it's true:) I've changed it:) Thank you!

package.json
VitalinaKuzmenko: I don't know. I was installing dependencies using npm install. And it was installed automatically. Is something wrong with it?

sampennington: So there is just one issue, eslint should be included in the dev dependencies section.

There are two types of dependencies:
- "dependencies": Packages required by your application in production. (production is just the live website you deploy)
- "devDependencies": Packages that are only needed for local development and testing.

Eslint falls into the second one

sampennington: You can read more about this here: https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

VitalinaKuzmenko: But I have it inside devDependencies, no? 
"devDependencies": {
    "eslint-plugin-react-hooks": "^4.6.0",

Maybe I will ask you personally about it :) Thank you!

PR #542, opened 2023-02-28
London 9 Turing Khadija Abdurahman Haji - React Week 1

KhadijaHaji:


src/App.js
```diff
@@ -1,13 +1,28 @@
...
 
+// what is key about...
+
+
 const App = () => {
+  const contact = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789",
+  ];
```
sampennington, line 18: It could better to define this variable outside of the app component. Like this we are recreating it every time the App component rerenders and we dont need to 😄

src/App.js
```diff
@@ -1,13 +1,28 @@
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards />
       <Bookings />
+      <Restaurant />
+      <Footer contact={contact} />
```
sampennington, line 25: Love this structure, so easy to see whats doing what!

src/Bookings.js
```diff
@@ -1,18 +1,54 @@
...
+      if (res.ok) {
+        return res.json().then((data) => {
+          setBookings(data);
+          setLoading(false);
+        });
+      } else {
+        setErrorOccurred(true);
+      }
+    });
+  }, []);
```
sampennington, line 35: There is another way that this fetch function might fail, do you know what it is? 😄

src/CustomerProfile.jsx
```diff
@@ -0,0 +1,28 @@
...
+            <p>VIP: {profile.vip ? "yes" : "no"}</p>
+            <p>Customer Phone Number: {profile.phoneNumber}</p>
+        
+        </div>
+        ) 
+      
+    )
+   
+}
+
```
sampennington, line 27: This component is a little hard to read, I think you would benefit from getting your linting setup with autofix on.
Going to try and schedule some help around this!

src/Footer.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = (props) =>{
+ return (
+   <ul className="contact-container">
+     {props.contact.map((lineString) => {
```
sampennington, line 6: `lineString` is an odd variable name 🤔  ~Could we think of anything better to use?

src/Heading.jsx
```diff
@@ -0,0 +1,12 @@
...
+
+const Heading = () => {
+return (
+  <header className="App-header">
+    <h1>CYF Hotel</h1>,
+    <img src="https://image.flaticon.com/icons/svg/139/139899.svg" alt="icon" />
+  </header>
+);
+}
+
```
sampennington, line 11: Love the small compoents, so easy to see what is doing what! 😄

src/Restaurant.js
```diff
@@ -1,14 +1,33 @@
...
       <ul>
-        <li>
-          Pizzas: {pizzas} <button className="btn btn-primary">Add</button>
-        </li>
+        <Order orderType="pizzas" />
+        <Order orderType="salads" />
+        <Order orderType="chocolate" />
       </ul>
     </div>
   );
```
sampennington, line 33: Generally its better to split your components into seperate files, it is easier to see whats going on 😄

src/SearchResults.jsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from "react";
+import moment from "moment";
+import CustomerProfile from "./CustomerProfile";
+import NewBookings from "./NewBookings";
+
+
+const getDaysDifference = (date1, date2) => {
+  let a = moment(date1);
+  let b = moment(date2);
```
sampennington, line 9: Think we can use `const` here? Always better to use const when we are not going to be reassigning a variable 😄

src/SearchResults.jsx
```diff
@@ -0,0 +1,70 @@
...
+  return a.diff(b, 'days')
+
+}
+
+const SearchResults = (props) => {
+  const [selectedId, setselectedId] = useState(undefined);
+  const [clickIdProfile, setClickIdProfile] = useState(undefined);
```
sampennington, line 16: You dont actually need to put `undefined` into the `useState()` here, it is this by default

src/TouristInfoCards.jsx
```diff
@@ -0,0 +1,48 @@
...
+          className="card-img-top"
+        />
+        <div className="card-body">
+          <a href="https://www.visitlondon.com/" className="btn btn-primary">
+            More Information
+          </a>
+        </div>
+      </div>
+    </div>
+  );
```
sampennington, line 45: There is some duplication here, what can we do about that ? 😄

sampennington (commented):
Great work Khadija! 
So happy to see the progress you are making 🥳 
Have left a few comments, no major issues! But I would suggest maybe trying to delete your comments, and keeping components in seperate files

PR #544, opened 2023-02-28
London Class 9 - Leila Farsani - React 

leilafarsani:
Live Demo : https://cyf-leilafarsani-hotel-react-app.netlify.app


src/App.js
```diff
@@ -1,13 +1,22 @@
...
-
-import Bookings from "./Bookings";
+import Heading from "./Heading.js";
+import TouristInfoCards from "./TouristInfoCards.js";
+import Bookings from "./Bookings.js";
+import Footer from "./Footer.js";
 import "./App.css";
 
 const App = () => {
+  const footerArray = [
```
jonnywyatt, line 9: This works fine; because the links are unlikely to change often, maybe simpler just to put them inside the `Footer` component instead of passing them in as a prop.

src/App.css
```diff
@@ -49,9 +43,18 @@ tr {
...
   text-align: center;
+  list-style-position: inside;
+  background-color: #0d0980;
+  color: white;
 }
 
 .card {
   width: 18rem;
 }
+.cards-box {
```
jonnywyatt, line 56: To think of this as 'mobile first', imagine how it would best look on a small screen first (probably, the 3 cards stacked vertically). Write the styles for that, then add a media query for eg min-width: 768px and adjust the styles inside the media query for the wider screen. If you can get in the habit of writing layout styles mobile-first, it will become simpler to approach it each time, and it will work on all screen sizes.

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
...
+  "Email",
+  "room Id",
+  "Check-In",
+  "Check-Out",
+  "Night"
+];
+const SearchResults = props => {
+  return (
+    <table className="table">
+      <thead>
```
jonnywyatt, line 18: Good table HTML 👏 
Later, we can look at how to add a <caption /> and hide it except for screen reader users, if required.
Also, big tables with lots of columns are hard to fit on small screens. One technique is to hide some columns on small screens; another is to enable the table to scroll sideways.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,48 @@
...
+        "Glasgow is a city in Scotland, located in the west central lowlands of the country. It is the largest city in Scotland and the fourth largest in the United Kingdom, with a population of over 600,000 people."
+    }
+  ];
+  return (
+    <div className="cards-box">
+      {cities.map(({ image, name, summary, website }) => (
+        <div className="card">
+          <img src={image} className="card-img-top" alt={name} />
+          <div className="card-body">
+            <h3>{name}</h3>
```
jonnywyatt, line 36: Maybe `h2` is best (also the page needs an `h1`)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,48 @@
...
+            <p>{summary}</p>
+            <a href={website} className="btn btn-primary">
```
jonnywyatt, line 38: Great, even though it looks like a button, it's a link so should be an `a`

jonnywyatt (approved):
Great job! 
Some tips to organise your CSS -
- separate the styling of 'common' things eg font face, header, footer, buttons at the top of the file. Later it will be easier to extract them to a common stylesheet for use on all pages
- use `rem` for font sizes
- use CSS vars for colours, and put them at the top of your common file
- use a CSS var for a spacing unit eg `$spacing_unit: 10px` and use multiples of it for your spacing. It will give the whole layout more consistency. eg `margin-bottom: $spacing_unit * 3`
- don't nudge left or right margins with px, instead align elements if you can, it will make responsive layout easier
- write your styles 'mobile-first'

PR #545, opened 2023-03-01
London-9Lovelace-Boshra-Mahmoudi-hotel-react

BoshraM:


src/App.css
```diff
@@ -50,8 +85,47 @@ tr {
...
+  align-self: center;
+}
+
+h3 {
+  text-align: center;
+}
+
+.restaurant-list {
+  display: grid;
+  grid-template-columns: 1fr auto auto;
```
sampennington, line 152: Love the use of grid 🥳

src/App.js
```diff
@@ -1,13 +1,54 @@
...
+const Manchester = {
+  src:
+    "https://eu-assets.simpleview-europe.com/manchester2016/imageresizer/?image=%2Fdbimgs%2FManOne-03%284%29.jpg&action=Background_Overlay",
+  cardTitle: "Manchester",
+  cardText:
+    " Manchester is the only UK city to feature in Lonely Planet's Best in Travel 2023 list and the only UK city in National Geographic's influential ‘Best of the World’ list which annually sets out 25 of the must-see places to visit around the globe.",
+  href: "https://www.visitmanchester.com/"
+};
+
+const London = {
```
sampennington, line 27: I like how you are constructing the data here! But I would still use a lower case for all js variables like this. Even though London is a name of a place etc!

src/Bookings.js
```diff
@@ -1,18 +1,47 @@
...
+  const [searchVal, setSearchVal] = useState("");
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState(false);
+
+  // let booking = () =>{
+  //   setBookings(bookings.concat(FakeBookings));
+  // }
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/delayed`).then(res => {
+      if (res.ok) {
```
sampennington: Hhandling the errors, nice! However theres another possibility here, do you know what it is?

The fetch function can actually throw an error too in some cases, what should we do about it? 😄

src/CustomerProfile.js
```diff
@@ -0,0 +1,26 @@
...
+
+const CostomerProfile = ({ customerProfile }) => {
+  const [profile, setProfile] = useState("");
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${customerProfile}`)
+      .then(res => res.json())
+      .then(data => {
+        setProfile(data);
+      });
```
sampennington, line 11: Any error handling we could do here? Like above!

src/Footer.js
```diff
@@ -0,0 +1,17 @@
...
+      <ul>
+        {props.adress.map((item, key) => (
+          <li className="footer-list" key={key}>
+            {item}
+          </li>
+        ))}
+      </ul>
+    </div>
+  );
+};
```
sampennington, line 15: Loving the small components, so easy to see what everything does!

src/Order.js
```diff
@@ -0,0 +1,23 @@
...
+  const orderOne = () => {
+    setOrders(orders + 1);
+  };
+
+  const orderLess = () => {
+    orders > 0 ? setOrders(orders - 1) : setOrders(0);
+  };
+
+  return (
+    <>
```
sampennington, line 16: We don't need to use this <></> here actually, this is only one element 😄

src/RestaurantButton.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+export const RestaurantButton = props => {
+  return (
+    <>
```
sampennington, line 5: But here we do need to use the fragments, nice! Good idea to use this than another DOM element 👍

src/SearchResults.js
```diff
@@ -0,0 +1,69 @@
...
+            <th scope="col">room id</th>
+            <th scope="col">check in date</th>
+            <th scope="col">check out date</th>
+            <th scope="col">nights stayed</th>
+          </tr>
+        </thead>
+        <tbody>
+          {props.data.map(person => {
+            let a = person.checkOutDate;
+            let b = person.checkInDate;
```
sampennington, line 36: Maybe we could use `const` for these variables? It doesn't look like they change

src/TouristInfoCards.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+
+const TouristInfoCards = props => {
+  return (
+    <div className="card-container">
+      {props.cities.map((city, key) => (
+        <div className="card" key={key}>
```
sampennington, line 7: It is often best to not use the array index as the key here. 
Take a look at the keys sections in the docs, there is a better explanation than I could ever give 😊  https://legacy.reactjs.org/docs/lists-and-keys.html

sampennington (commented):
Great stuff Boshra!! Love to see how far we have come 😄

PR #546, opened 2023-03-01
London 9 - Andriana Saffo - React - Week 1

AndrianaOS:


src/App.js
```diff
@@ -1,13 +1,64 @@
...
+import Footer from "./Footer";
+import Restaurant from "./Restaurant";
+
+const info = [
+  "123 Fake Street, London, E1 4UD",
+  "hello@fakehotel.com",
+  "0123 456789"
+];
+
+const TouristInfoDetails = [
```
sampennington, line 15: I like that we are separating our data from our components, makes the components a lot simpler!

But it is good to use camelCase for our JS variables! This is starting with a capital

src/App.js
```diff
@@ -1,13 +1,64 @@
...
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <div className="card-content">
+        {TouristInfoDetails.map((detail, index) => {
+          return (
+            <TouristInfoCards
+              key={index}
```
sampennington, line 50: Its best to try and avoid using the array index as the key here. Have a read of the docs here: https://legacy.reactjs.org/docs/lists-and-keys.html#keys
That can explain it better than I can! 😄

src/BookingForm.js
```diff
@@ -0,0 +1,135 @@
...
+    } else if (event.target.name === "email") {
+      setEmail(event.target.value);
+    } else if (event.target.name === "roomId") {
+      setRoomId(event.target.value);
+    } else if (event.target.name === "checkInDate") {
+      setCheckInDate(event.target.value);
+    } else if (event.target.name === "checkOutDate") {
+      setCheckOutDate(event.target.value);
+    }
+  }
```
sampennington: There is a way we can change our state here so we don't have all the `else if` statements, can you see what it is? 😄 
Hint: It would involve just having one `useState`

src/BookingForm.js
```diff
@@ -0,0 +1,135 @@
...
+
+  // console.log(props.results);
+
+  function handleSubmit(event) {
+    event.preventDefault();
+    let randomID = props.results.length + 1;
```
sampennington: It looks like we could use `const` here

src/Bookings.js
```diff
@@ -1,21 +1,140 @@
...
+
+  function handleAddPerson(newBooking) {
+    setBookings([...bookings, newBooking]);
+  }
+
+  // console.log(bookings);
+
   const search = searchVal => {
     console.info("TO DO!", searchVal);
+    const filteredValue = bookings.filter(value => {
```
sampennington, line 25: Instead of `value` here, it could be good to use `booking`, it helps with readability, `value` is quite unclear what it actually is, but we know it is a `booking` from our `bookings` array!

src/Bookings.js
```diff
@@ -1,21 +1,140 @@
...
+        // console.log(data);
+        if (data.error) {
+          setHandleError(data.error);
+        } else {
+          setBookings(data);
+        }
+        setLoading(false);
+      })
+      .catch(error => {
+        console.log(error);
```
sampennington, line 50: Anything else we might want in our catch function here? We are only logging the error

src/Bookings.js
```diff
@@ -1,21 +1,140 @@
...
+      .then(data => {
+        // console.log(data);
+        if (data.error) {
+          setHandleError(data.error);
+        } else {
+          setBookings(data);
+        }
+        setLoading(false);
+      })
+      .catch(error => {
```
sampennington, line 49: This is great Adrianna! You have found that the `fetch` function can error in two ways, it can return with an error, but it can also throw an error!

src/SearchResults.jsx
```diff
@@ -0,0 +1,113 @@
...
+import CustomerProfile from "./CustomerProfile";
+
+function SearchResults(props) {
+  const [profile, setProfile] = useState(null);
+  const [sort, setSort] = useState("DEFAULT");
+  const [sortedResults, setSortedResults] = useState(props.results);
+
+  useEffect(() => {
+    console.log(sort);
+    let updatedResults = [...props.results].sort((a, b) => {
```
sampennington, line 12: Think we can use a `const` here? 😄

src/SearchResults.jsx
```diff
@@ -0,0 +1,113 @@
...
+        }
+      } else if (sort === "DESC") {
+        if (a.firstName > b.firstName) {
+          return 1;
+        } else if (a.firstName < b.firstName) {
+          return -1;
+        }
+      } else {
+        return 0;
+      }
```
sampennington, line 27: There is a lot of logic here, to make it a bit clearer it might be a good idea to separate this into a separate function

src/TouristInfoCards.js
```diff
@@ -0,0 +1,18 @@
...
+      <img src={cities.pic} className="card-img-top" />
+      <section className="card-body">
+        <h3>{cities.name}</h3>
+        <p>{cities.description}</p>
+        <a href={cities.url} className="btn btn-primary">
+          More about {cities.name}
+        </a>
+      </section>
+    </div>
+  );
```
sampennington, line 15: Nice simple componets, so easy to see what is going on ❤️

sampennington (commented):
Looking good Andrianna! Few little bits, if you don't know what I mean by anything reach out on slack!

PR #547, opened 2023-03-02
london-9-cyf-hotel-react

laxmikckarki:
laxmi karki kc


src/App.js
```diff
@@ -1,13 +1,34 @@
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <div className='cards-container'>
+          <TouristInfoCard image="Glasgow.jpeg" title="Glasgow" description="Glasgow is a port city on the River Clyde in Scotland's western
+        Lowlands. It's famed for its Victorian and art nouveau architecture..." link="https://peoplemakeglasgow.com"/> 
+          <TouristInfoCard image="London.jpg" title="London" description="London, the capital of England and the United Kingdom, is a
+         21st-century city with history stretching..." link="https://visitlondon.com"/> 
+          <TouristInfoCard image="Manchester.jpg" title="Manchester" description="Manchester is a major city in the northwest of England with
```
sampennington, line 26: Nice laxmi!!, you have really understood using components to remove duplication!! You are the first ive seen to use a separate card component for each card here 😄

src/Order.js
```diff
@@ -0,0 +1,19 @@
...
+return(
+<li>
+    {props.orderType}: {orders} <RestaurantButton order = {orderOne}/> 
+</li>
+);
+}
+
+
+
+
```
sampennington: Not jugley important but its better to not leave lots of empty lines! One is the most you should leave really to break things up 😄

src/RestaurantButton.js
```diff
@@ -0,0 +1,8 @@
+import React, { useState } from "react";
+
+const RestaurantButton = (props) => {
+    return 
+    <button className="btn btn-primary" onClick={props.order}>Add</button>
```
sampennington: Could we put the button on the same line as our return statment? Or we could wrap the button in brackets ().

It looks a bit like this is returning nothing otherwise!

sampennington (commented):
Loooking greta Laxmi just left a couple comments as I saw this 😄

src/App.js
laxmikckarki: thank you sam for your beautiful comment.

src/Order.js
laxmikckarki: yes, you are right. i will consider next time and correct this.

src/RestaurantButton.js
laxmikckarki: yes, can do that.

PR #548, opened 2023-03-03
London9-Elahe-Mortazavi-cyf-hotel-react

elahemortazavi:


src/Order.js
```diff
@@ -0,0 +1,28 @@
...
+
+const Order = ({orderType}) => {
+
+    const [order, setOrder] = useState(0);
+
+    const clickHandler = () => {
+        setOrder(order + 1);
+    }
+
+    const clickRemover = () => {
```
sampennington: We might be able to think of a better names for these `clickHandler` and `clickRemover` 🤔 

Maybe we could relate there name to the state they update? Something like, `addOrder`?

src/SearchResults.js
```diff
@@ -0,0 +1,51 @@
...
+import moment from "moment";
+
+const SearchResults = (props) => (
+  <table className="table table-striped">
+    <CreateTableHeader />
+    <CreateEachRow results={fakeBookings} />
+  </table>
+);
+
+const CreateTableHeader = () => (
```
sampennington: Loving the small components!! So easy to see whats going on 😄 

I would only just suggest that you don't need to add the prefix `Create`. 
You're right, this does create a table header, but its not the convention to add it. most react components create some html, but its unnecessary to specify I think

PR #549, opened 2023-03-03
London-9-Turing-Farnoosh_Moayeri-React-WEEK1

Farnooshmo:
**CYF [Hotel**](https://cyf-farnooshmo-hotel-react.netlify.app/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,9 +14,26 @@ const App = () => {
   return (
     <div className="App">
       <Heading />
-      <TouristInfoCards />
+      <TouristInfoCards
+        image=""
+        title="Glasgow"
+        link="href=https://peoplemakeglasgow.com"
+        description="Is the most populous city in Scotland and the fourth-most populous city in the United Kingdom, as well as being the 27th largest city by population in Europe"
+      />
+      <TouristInfoCards
+        image=""
+        title="Manchester"
+        link="href=https://visitmanchester.com"
+        description="Is a major city in the northwest of England with a rich industrial heritage. The Castlefield conservation area’s 18th-century canal system recalls the city’s days as a textile powerhouse, and visitors can trace this history at the interactive Museum of Science & Industry."
+      />
+      <TouristInfoCards
+        image=""
+        title="London"
+        link="href=https://visitlondon.com"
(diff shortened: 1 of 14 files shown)


Farnooshmo:
@qingwei91 Hi, could you please review my code?

src/App.js
```diff
@@ -1,13 +1,44 @@
...
+  const contact = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789",
+  ];
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <TouristInfoCards
```
qingwei91: small improvement: you can store these data in a data structure like a list of object, and then produce List of `TouristInfoCards` from that.

This approach will be more flexible, and less verbose

src/Bookings.js
```diff
@@ -1,18 +1,39 @@
...
+      .then((res) => res.json())
+      .then((data) => {
+        setBookings(data);
+      });
+  }, []);
+  const search = (searchVal) => {
     console.info("TO DO!", searchVal);
+    const filteredBooking = bookings.filter((name) =>
+      name.firstName.toLowerCase() === searchVal.toLowerCase() ||
+      name.surname.toLowerCase() === searchVal.toLowerCase() ||
```
qingwei91: I think the top 2 conidition might be redundant since you're performing includes search next?

src/Bookings.js
```diff
@@ -1,18 +1,39 @@
...
+      name.firstName.toLowerCase().includes(searchVal.toLowerCase()) ||
+      name.surname.toLowerCase().includes(searchVal.toLowerCase())
+        ? setBookings([name])
```
qingwei91: If you're performing some state changes, it might be better to call setBookings outside of filter function.

Generally, filter function should be pure, pure in this context means it does not mutate any variables that is outside of the scope of the function.

src/Bookings.js
Farnooshmo: You are right, they can be deleted.

Farnooshmo:
@qingwei91  Thank you for the review; I fixed them,  could you please review them?

PR #551, opened 2023-03-05
 Mickey Haile-london 9-Lovelace- cyf-hotel-react

mickeyhaile2:


src/Bookings.js
```diff
@@ -1,21 +1,65 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [hasError, setHasError] = useState(false);
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me`).then(res =>
+      res.json().then(data => {
+        console.log(data);
+        setBookings(data);
+      })
+    );
+  }, []);
```
JDysiewicz, line 20: Really nice - as a small improvement you can actually extract that `const [bookings, setBookings] = useState([]);useEffect(() => { fetch(...) }, [])` logic into a custom hook for extra style points ;)

src/Bookings.js
```diff
@@ -1,21 +1,65 @@
...
+    <div className="container">
+      <Search search={search} />
+      {hasError ? (
+        <div>Oh no something went wrong!</div>
+      ) : isLoading ? (
+        <div>Loading...</div>
+      ) : (
+        <SearchResults results={bookings} />
+      )}
     </div>
```
JDysiewicz, line 54: Nice use of conditional rendering here - although careful when you start chaining 2,3,4.. ternarys in a row as it can be difficult to read. It's fine here, although personally I'd prefer these to be outside of the main return of the function (e.g. 
```
const Bookings = () => {
   ...

   if (hasError) {
      return <div>Oh no something went wrong!</div>
   }

   if (loading){
      return <div>Loading...</div>
   }

   return (
      <SearchResults />
   )
}
```

src/CustomerProfile.js
```diff
@@ -0,0 +1,28 @@
...
+
+const CustomerProfile = (props) => {
+  const [customerProfileData, setCustomerProfileData] = useState("");
+
+   useEffect(() => {
+     fetch(`https://cyf-react.glitch.me/customers/${props.customerId}`)
+       .then((res) => res.json())
+       .then((data) => setCustomerData(data))
+       .catch((err) => console.log(err));
+   }, [props.customerId]);
```
JDysiewicz, line 11: nice, you're getting the hang of this data fetching pattern!

src/CustomerProfile.js
```diff
@@ -0,0 +1,28 @@
...
+      <p>Customer {props.customerId} Profile</p>
+      <ul>
+        <li>Customer ID:{customerData.id}</li>
+        <li>Vip:{customerData.vip ? "Yes" : "no"}</li>
+        <li>Customer Email:{customerData.email}</li>
+        <li>Customer Phone Number:{customerData.phoneNumber}</li>
+      </ul>
+    </div>
+  ) : (
+    <p />
```
JDysiewicz, line 24: I'd try and choose a default tag which is the same as the top-level tag as the other response - in this case a `<div>` instead of a `<p>`, just because an empty paragraph tag doesn't make much sense (especially for those with screen readers, etc)

src/Footer.js
```diff
@@ -0,0 +1,20 @@
...
+  const contact = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
+  return (<div>
+      <p>Contact Us</p>
+      <ul>
+        {contact.map((item, index) => {
+          return <li key={index}>{item}</li>;
```
JDysiewicz, line 13: nice use of array methods and rendering components out

src/RestaurantButton.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+function RestaurantButton(props) {
+  return (
+    <div>
+      <button className="btn btn-primary" onClick={props.orderone}>
+        Add
+      </button>
+    </div>
```
JDysiewicz, line 9: The `<div>` here is unnecessary as you're only returning a single element

src/SearchResults.js
```diff
@@ -0,0 +1,66 @@
...
+                    booking.checkInDate,
+                    "days"
+                  )}
+                </td>
+                <td>
+                  <button className="btn btn-primary" onClick={handleClick}>
+                    Show Profile
+                  </button>
+                </td>
+              </tr>
```
JDysiewicz, line 58: This could be a separate `<TableRow>` component just to break it up a little more

JDysiewicz (commented):
A really solid piece of work, good job!

PR #554, opened 2023-03-11
London 9 Turing - Zahraa Tayyar - React - Week 1 + 2 + 3

zahraateee:


src/Bookings.js
```diff
@@ -1,18 +1,58 @@
...
+        }
+      })
+      .then((data) => {
+        // console.log("data", data);
+        setLoading(false);
+        setError(false);
+        setBookings(data);
+      })
+      .catch((error) => {
+        console.log("3wq ZV");
```
ehwus, line 36: Is this cthulu ?

src/Bookings.js
```diff
@@ -1,18 +1,58 @@
...
+      });
+  }, []);
+
+  const search = (searchVal) => {
     console.info("TO DO!", searchVal);
+    setBookings((prev) =>
```
ehwus, line 42: If we filter here based off the previous state value, then every search will have fewer and fewer options to actually grab. If I search and have no results, and then search again for someone who exists, then I wouldn't successfully find them.

To solve this, the most obvious option would be to store two arrays in state - listOfClients and filteredListOfClients maybe. We could simplify this futher though by just storing the actual search term in state as a string and then filtering the results based off of that.

ehwus (commented):
This is really solid! Everything functions well in terms of the brief, and it's great that it's deployed as well.

I think some of the ideas in your design are really good but could use a little bit of refinement to really make the whole thing shine, for instance:

Screenshot 2023-04-06 at 09 24 23

Your hero image (see above) is probably a bit too big and has led to some repetition on a bigger monitor - I'd make this element smaller and try to make sure that it scales well to multiple screen resolutions. It looks great on mobile though, and in general you've used responsive design really well in this project and it looks thoughtfully designed.

Where you haven't got this control over screen size, sometimes the image can be in a position that makes the white text slightly harder to read as well, so setting minimum and maximum sizes of the element and making sure that it's readable across a range of screens is an important part of the process. 

The code reads well and demonstrates a good level of understanding of everything we have covered on the course! I think that revising the searching again would be a useful exercise; filtering lists based on, for example, a search term is a pattern that is used quite a lot in React.

PR #557, opened 2023-03-16
London 9-Karleen Richards-cyf-hotel-react

karleenmsrichards:


src/App.css
```diff
@@ -2,43 +2,284 @@
...
 .App-header {
-  background-color: #222;
-  height: 50px;
-  padding: 20px;
-  color: white;
+  z-index: 500;
+  display: grid;
+  grid-template-columns: auto 0.5fr 0.5fr;
+  grid-template-areas: "heading . find guests account";
+  grid-template-columns: 1fr 0.5fr 0.7fr 0.7fr 0.2fr;
```
sampennington, line 10: Love that we're using grid, I need to learn this myself! 😂

src/App.css
```diff
@@ -2,43 +2,284 @@
...
+  padding: 2px 5px;
+  border-radius: 25px;
+  border: 1px solid rgb(237, 229, 230);
+}
+
+.Account:hover {
+  cursor: pointer;
+  box-shadow: 0px 15px 10px -15px rgb(103, 101, 101);
+  background-color: white;
+}
```
sampennington, line 86: Love that we are adding effects, so coool!!

src/Bookings.js
```diff
@@ -1,20 +1,64 @@
...
+        return response.json();
+      })
+      .then(data => {
+        console.log(data);
+        setErrorMessage(data.error);
+        setBookings(data);
+      })
+      .catch(error => {
+        setErrorMessage(error);
+      })
```
sampennington, line 39: This is great Karleen!! You have found that there are multiple ways in which the `fetch` function can error. It can be not `ok` and it can actually throw an error too! Do you know when each one might happen?

src/CustomerProfile.js
```diff
@@ -0,0 +1,39 @@
+import { useState, useEffect } from "react";
+
+export const CustomerProfile = ({ id }) => {
+  let [customerInfo, setCustomerInfo] = useState(null);
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then(response => response.json())
+      .then(data => {
```
sampennington, line 9: Have we considered all the ways this api could error? 😄

src/CustomerProfile.js
```diff
@@ -0,0 +1,39 @@
...
+        <li>VIP Customer</li>
+      ) : (
+        <li>Not a VIP Customer</li>
+      )}
+      <li>Customer Phone Number: {customerInfo.phoneNumber}</li>
+    </ul>
+  ) : (
+    <></>
+  );
+};
```
sampennington, line 32: While this is great! My preference would have been to just do:
```
customerInfo && (
<ul className="customer-profile">
      <h4>Customer Profile</h4>
      <li>Customer Email: {customerInfo.email}</li>
      <li>Customer Id: {customerInfo.id}</li>
      {customerInfo.vip === true ? (
        <li>VIP Customer</li>
      ) : (
        <li>Not a VIP Customer</li>
      )}
      <li>Customer Phone Number: {customerInfo.phoneNumber}</li>
    </ul>)
```

We don't really care if they are undefined or null specifically, just that they are defined, with a value. So the `customerInfo &&` covers both of these options. We also don't really need to return the empty fragment either, but this again this is theoretically fine, and is just a  preference of mine on what looks neatest 😃

src/Footer.js
```diff
@@ -0,0 +1,28 @@
+import { List } from "./List";
+
+const Footer = () => {
+  const footerElements = [
+    "123 Fake Street, London, E1 4UD",
+    "hello@fakehotel.com",
+    "0123 456789"
+  ];
```
sampennington: It would be better to define this variable outside of the component, as this is just a constant. If we do it like this we recreate this unnecessarily on every render.

src/List.js
```diff
@@ -0,0 +1,9 @@
+export const List = ({ array }) => {
```
sampennington: I think a better prop name for this would be `items`, it is uncommon to see a data type as a prop name 🤔

src/List.js
```diff
@@ -0,0 +1,9 @@
...
+  return (
+    <ul className="footer-ul">
```
sampennington: This is a nice component that could be reused lots of places, but you have named it's className after the footer. Could it be more general maybe?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,30 @@
...
+export const TouristInfoCards = () => {
+  return (
+    <div className="cards-wrapper">
+      <Card
+        url="https://cdn.londonandpartners.com/-/media/images/london/visit/things-to-do/sightseeing/london-attractions/coca-cola-london-eye/the-london-eye-2-640x360.jpg?mw=640&hash=F7D574072DAD523443450DF57E3B91530064E4EE"
+        location="Lonon Eye"
+        city="London"
+        description="From the modern London Eye to the historic Tower of London, the top 10 tourist attractions in London are a must-see on any London sightseeing trip. Even better, many London landmarks are free to visit."
+        link="https://visitlondon.com"
+      />
```
sampennington, line 13: Great we have made a separate Card component to be used here! 😎

src/SearchResults.js
```diff
@@ -0,0 +1,90 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {results.map(obj => {
+            // let a = moment(obj.checkInDate, "YYYY-MM-DD");
+            // let b = moment(obj.checkOutDate, "YYYY-MM-DD");
+            const dateOne = new Date(obj.checkInDate);
+            const dateTwo = new Date(obj.checkOutDate);
+            const Difference_In_Time = dateTwo.getTime() - dateOne.getTime();
+            const Difference_In_Days = Difference_In_Time / (1000 * 3600 * 24);
```
sampennington: Any reason we have started using capital letters for variables, and under scores? Consistency is important, otherwise a codebase becomes a mess!
In JS we tend to juist stick with camelCase for variables, like you have everywhere else in this work 😄

src/Order.js
```diff
@@ -0,0 +1,21 @@
+import { useState } from "react";
+import { RestaurantButton } from "./RestaurantButton.js";
+
+export const Order = ({ orderType }) => {
+  let [orders, setOrders] = useState(0);
+
+  const OrderOne = () => {
```
sampennington: Why is this a capital letter? We should be using camel case for functions. With the exception of React components, who does require a capital letter.

sampennington (commented):
I love seeing how far you have come Karleen this is sooo cool! Just a few comments but overall it is amazing work ❤️

sampennington:
Oh one other thing! There is a lot of comments and console logs, its a lot neater if you tidy these up after, so try and remember to remove them when you are done 👍

src/CustomerProfile.js
karleenmsrichards: I have considered your comment and have used then instead of else and the empty fragment.

src/Footer.js
karleenmsrichards: Oh I see, sounds like  good idea. I took heed. thank you.

src/List.js
karleenmsrichards: Thanks for pointing that put. It would be a good decision to no have data types as prop names, thanks Sam

src/List.js
karleenmsrichards: Good catch, will come up with a more generic reusable one 😀

src/List.js
karleenmsrichards: Do you have any ideas? Coming to think of it I can only think of list or unordered-list as a class name, do they follow best practice?

src/SearchResults.js
karleenmsrichards: 😂 This is what you achieve for copying code lol... I believe I found this snippet on Stack Overflow, I should have amended it... I will now. thanks much.

src/Order.js
karleenmsrichards: I have no idea why it is, my error. I will make a change to it now.

src/App.css
karleenmsrichards: ☺️

src/Footer.js
karleenmsrichards: Ok I will.

PR #558, opened 2023-03-16
LONDON9-TURING-GAYLE_THOMPSON-IGWEBIKE-CYF_HOTEL-REACT-WK3

Gayle-Thompson-Igwebike:


src/App.css
```diff
@@ -1,21 +1,32 @@
...
 }
 
 .App-header {
-  background-color: #222;
-  height: 50px;
+  background-color: var(--yellow);
+  height: 100px;
   padding: 20px;
-  color: white;
+  color: var(--gray-dark);
```
sampennington, line 21: Nice using variables here!

src/App.css
```diff
@@ -47,11 +58,184 @@ tr {
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: center;
+  column-gap: 50px;
+  row-gap: 50px;
+  margin: 3rem;
+ }
+
+ .info-image{
```
sampennington, line 83: Try to leave a space between the  classname and the bracket if you can 😄

src/TouristInfoCards.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import HotelImage from "./images-folder/hotel.jpeg";
+
+function TouristInfoCards(props) {
```
sampennington, line 4: I love that we have broken this down into a small component! So easy to see what is going on 😄 
But we have called it Cards over Card would be my only issuee, there is only one card in here. Best to make it as clear as possible what a component is actually doing!

src/Bookings.js
```diff
@@ -1,21 +1,69 @@
...
+  
+  const search = (searchVal) => {
+    setTypedSearchValues(searchVal);
   };
 
+  function addNewBooking(newBooking){
+    setBookings([...bookings, newBooking])
+  }
+
+  const filteredBookings = bookings.filter((customerName) => {
```
sampennington, line 21: I don't think we have named this `customerName` variable very well 🤔 
What is this value actually? It's a `booking` from our `bookings` array! So we should call it `booking` 😎

src/Bookings.js
```diff
@@ -1,21 +1,69 @@
...
+          setErrorMessage(data.error);
+        } else {
+          setBookings(data);
+        }
+
+        setLoading(false);
+      })
+      .catch((error) => {
+        console.log(error)
+      });
```
sampennington, line 45: Love to see the error handling!
But what happens in the catch block here, do we not want to tell the user something went wrong?

src/Heading.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import logo from "./images-folder/black-cyf-logo.jpeg"
+
+
+function Heading() {
+  const logoimage = logo;
```
sampennington: I don't think theres any point in making a new variable here, you can just pass it straight to the `src` prop

src/NewBookingForm.js
```diff
@@ -0,0 +1,127 @@
...
+    } else if (event.target.name === "email") {
+      setEmail(event.target.value);
+    } else if (event.target.name === "roomId") {
+      setRoomID(event.target.value);
+    } else if (event.target.name === "checkInDate") {
+      setCheckInDate(event.target.value);
+    }else if (event.target.name === "checkOutDate") {
+      setCheckOutDate(event.target.value);
+    }
+}
```
sampennington, line 30: Can we think of a way to simplify this? There is a way of doing this without all of the `if/elses`.
It would involve keeping them all in one state object 😄  If you are not sure, then reach out to me I can help!

sampennington (commented):
Awesome work Gayle!! So happy to see how much progress you are making ❤️ 

Just a few notes, but also there are a lot of comments and consoles left around, always a good idea to clean them up when you are done 😄

src/App.css
Gayle-Thompson-Igwebike: Thank you Sam.

src/App.css
Gayle-Thompson-Igwebike: Gotcha. x

src/Bookings.js
Gayle-Thompson-Igwebike: I understand what you mean. I will update it once I can fetch the API again so I can test that it is working.... But I get your point here.

src/Bookings.js
Gayle-Thompson-Igwebike: I have updated it now to tell the user that "The requested information is not available at the moment"

src/Heading.js
Gayle-Thompson-Igwebike: Deleted and updated....Thank you.

src/NewBookingForm.js
Gayle-Thompson-Igwebike: Soooooo......yes please show me how to make it simpler. I did it like this because it is easier for me to understand what I am doing and trying to achieve. But always happy to learn a new way of doing things. Thanks Sam.xxx

sampennington: This is a thing a lot of people have done, so I am going to bring this to the help center! 😄

PR #561, opened 2023-03-17
London 9 - Lorena Capraru - React

LorenaCapraru:


src/App.css
```diff
@@ -7,15 +7,30 @@
...
-.App-header {
+.imgLogo {
+  height: 100px;
+}
+.cardWrapper {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+  flex-wrap: wrap;
+}
```
sampennington, line 18: Its a good idea to try and put a space in between these classes to seperate them if you can 😄

src/App.js
```diff
@@ -1,13 +1,52 @@
...
+      img: "https://www.propertyreporter.co.uk/images/660x350/manchester_895-14609.jpg",
+    },
+    {
+      name: "London",
+      description:
+        "What is London? London is the capital city of the United Kingdom.It is the U.K.'s largest metropolis and its economic, transportation, and cultural centre.",
+      link: "https://visitlondon.com",
+      img: "https://cdn.londonandpartners.com/-/media/images/london/visit/things-to-do/sightseeing/london-attractions/coca-cola-london-eye/the-london-eye-2-640x360.jpg?mw=640&hash=F7D574072DAD523443450DF57E3B91530064E4EE",
+    },
+  ];
```
sampennington, line 35: This is a great idea Lorena! To split your data out from the component, but it could be improved a bit. 
Right now it is taking up a lot of the App file, but it doesn't really need to. 
There is also an issue that we will recreate this variable every time we rerender this app compeonnt.

So it would be better to define the variable outside of the component, maybe even in another file.

Same with the footer adress below, we don't need to define that inside our component 😄

src/App.js
```diff
@@ -1,13 +1,52 @@
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading />
+      <Hero />
+      <TouristInfoCards card={cards} />
       <Bookings />
+      <Restaurant />
+      <Footer address={footerAddress} />
```
sampennington, line 49: This looks very neat, I love it!

src/Bookings.js
```diff
@@ -1,18 +1,55 @@
...
+      })
+      .then((data) => {
+        setTimeout(() => {
+          setBookings(data);
+          setLoading(false);
+        }, 1000);
+      })
+      .catch((err) => {
+        setError("Error!");
+        setLoading(false);
```
sampennington, line 40: This is great Lorena! You have handled all the errors well. 😊

src/Bookings.js
```diff
@@ -1,18 +1,55 @@
...
+        booking.surname.toLowerCase().includes(searchVal)
+      );
+    });
+    setBookings(filterBookings);
   };
+  const [bookings, setBookings] = useState([]);
+
+  const [loading, setLoading] = useState(true);
+
+  const [error, setError] = useState(null);
```
sampennington, line 22: Best to keep your hooks at the very top of the file, so these three `useState` hooks would be better defined above the `search` function.

src/Hero.jsx
```diff
@@ -0,0 +1,35 @@
...
+          <i>
+            Our experienced and friendly team will ensure that your special day
+            is detailed, fabulous and forever memorable.
+          </i>
+        </p>
+      </div>
+      <hr />
+    </div>
+  );
+}
```
sampennington, line 35: Lots of simple components, this is a great way to do it, as its so easy to see whats going on!

src/SearchResults.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function SearchResults(props) {
+  const [selectedRow, setSelectedRow] = useState(null);
+  function selectRow(id) {
+    id === selectedRow
+      ? setSelectedRow(null) && setCustomerProfile(null)
+      : setSelectedRow(id);
+  }
+
+  const [customerProfile, setCustomerProfile] = useState(null);
```
sampennington, line 13: Again best to move this up to the top of the component!

src/SearchResults.jsx
```diff
@@ -0,0 +1,67 @@
...
+  useEffect(() => {
+    if (selectRow) {
+      fetch(`https://cyf-react.glitch.me/customers/${selectedRow}`)
+        .then((res) => res.json())
+        .then((data) => setCustomerProfile(data));
+    }
```
sampennington, line 19: Could this one error too?

src/TouristInfoCards.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+export default function TouristInfoCards(props) {
+  let card = props.card.map((el) => (
```
sampennington, line 4: Think this could be a `const`? Always better to use const where you dont need to  reassign a variable 😄

sampennington (commented):
Awesome work Lorena!! If you could just try and clean up comments it would be even neater!!

PR #573, opened 2023-05-31
Glasgow class 6 MeleseBerehannu hotel react Exercise 6

Melesegithub:
I have tried the rest as well please have a look the assigned part

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,21 @@
+import SearchResults from './SearchResults';
+import  Footer from './Footer';
+import TouristInfoCards  from './TouristInfoCards';
+import  Heading  from './Heading';
 import React from "react";
 
 import Bookings from "./Bookings";
 import "./App.css";
 
+
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading     />
       <Bookings />
+      
+      <TouristInfoCards     />
+      <Footer     />
     </div>
   );
 };
--- a/src/Bookings.js
+++ b/src/Bookings.js
@@ -1,8 +1,9 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
 
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
+export const results={FakeBookings}
 const Bookings = () => {
   const search = searchVal => {
     console.info("TO DO!", searchVal);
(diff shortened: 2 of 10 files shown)


src/App.js
```diff
@@ -1,13 +1,21 @@
...
 
 import Bookings from "./Bookings";
 import "./App.css";
 
+
 const App = () => {
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
+      <Heading     />
```
hanl00, line 14:
```suggestion
      <Heading />
```

you dont need a tab, just a regualar space would be enough

same for 

````
      <TouristInfoCards     />
      <Footer     />
````

File: src/Bookings.js [hunk 2]
```diff
@@ -1,8 +1,9 @@
 import React from "react";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
 
+import SearchResults from "./SearchResults.js";
+import FakeBookings from "./data/fakeBookings.json";
+export const results={FakeBookings}
```
hanl00, line 6: - this should be export default Bookings as you are importing Bookings on Line 7 in src/App.js,
` import Bookings from "./Bookings";`
- usually we tend to have default exports at the bottom of the file
```
const Bookings = () => {
  const search = searchVal => {
  };

  return (
    <div className="App-content">
      <div className="container">
        <Search search={search} />
        <SearchResults results={FakeBookings} /> 
      </div>
    </div>
  );
  
  export default Bookings
```

src/Footer.js
```diff
@@ -0,0 +1,18 @@
...
+  const arrayElement = array.map((element) => (
+    <li key={array.indexOf(element)}>{element} </li>
+  ));
+
+  return (
+    <footer>
+      <ul>{arrayElement} </ul>
+    </footer>
+  );
+}
```
hanl00, line 17: consider sticking to 1 way to declare React Functional Components for consistency

either
```
const MyComponentName = () => {
}
```
or 
```
function MyComponentName({ }) {
}
```

hanl00 (commented):
Looks good, seems like you have a decent understanding of React so far. 

My comments are all relating to code readability, consistency and formatting. Which can be easily fixed by adding a code formatter (and a linter too)

I would suggest Prettier and ESLint as they are the most widely used ones

https://dev.to/knowankit/setup-eslint-and-prettier-in-react-app-357b
https://prettier.io/

Prettier on vscode

https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

src/SearchResults.js
```diff
@@ -0,0 +1,39 @@
...
+      <th scope="col">Email</th>
+      <th scope="col">Room Id</th>
+      <th scope="col">Check in date</th>
+      <th scope="col">Check out date</th>
+      </tr>
+  </thead> 
+  <tbody>
+    
+    {results.map((element)=>{
+     return <><tr><td>{element.id}</td>
```
hanl00, line 24: you can use object destructuring here instead of calling element every line

PR #575, opened 2023-06-06
Glasgow - Class 06 - Malkit Benning - Hotel Project up step 22

malkitbenning:
Joint Exercise - students:
Malkit, Joe, Delnia, Siver


src/App.js
```diff
@@ -1,13 +1,27 @@
...
       <Bookings />
+
+      <TouristInfoCards />
+      <Footer
+        addressDetail={[
+          "123 Fake Street, London, E1 4UD",
+          "hello@fakehotel.com",
+          "0123 456789",
+        ]}
+      />
```
hanl00, line 23: Although this code works, on a readability note this bit could be improved

Perhaps renaming `addressDetail `to something like `contactDetails`


Or something like

```
<Footer
        address="123 Fake Street, London, E1 4UD"
        email="hello@fakehotel.com"
        phoneNumber="0123 456789"
 />
```

src/Footer.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+const Footer = (props) => (
+  <ul>
+    {props.addressDetail.map((addressLine, index) => {
+      return (
+        <li key={index}>{addressLine}</li>
+      )
+    })}
```
hanl00, line 9: if the above changes are applied then the following would look something like this
```suggestion
         <li key={address}>{address}</li>
         <li key={email}>{email}</li>
         <li key={phoneNumber}>{phoneNumber}</li>
```

also I've replace the list key to be the values instead of an index. Generally React does not recommend using index as keys, but i know this has been covered in the course so far. Just something to keep in mind when working in React Projects outside of CYF

https://dev.to/shiv1998/why-not-to-use-index-as-key-in-react-lists-practical-example-3e66

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "moment": "^2.29.4",
     "react": "^18.2.0",
```
hanl00, line 7: you are using React 18 here so you don't have to do `import React from "react";` on top of every file unless you are using something from react like hooks. I know that `import React from "react";` is provided by the coursework template so far, just something for you to keep in mind for future projects

https://stackoverflow.com/questions/71725865/do-i-need-to-import-react-in-every-file

src/TouristInfoCards.js
```diff
@@ -0,0 +1,62 @@
...
+      <div className="card-body">
+        <h5 className="card-title">{title}</h5>
+        <p className="card-text">{description}</p>
+        <a href={link} className="btn btn-primary">
+          More Information
+        </a>
+      </div>
+    </div>
+  );
+};
```
hanl00, line 17: This file src/TouristInfoCards.js contains 2  components, a Card component and a TouristInfoCards component

Generally it is best practice to have 1 component per file, https://github.com/airbnb/javascript/blob/master/react/README.md#basic-rules. 

Your Card compoent is a Pure Stateless component, so it is totally ok to have 2 components in 1 file here 

But it would still have it in `src/card ` because the Card component could be reused by other new components. Imagine that you want to create another React Component using the same Card component you have defined here.

Problem no.1 the card isn't being exported right now. If it was exported, all Card component imports would look like this

`import {Card} from ./TouristInfoCard` --- which doesn't sound right and can get real confusing if you do this on multiple components across multiple files on a larger codebase

instead of 

`import Card from ./Card`

Problem no.2 you would want to isolate each component in a single test file

hanl00 (commented):
Looks good, just minor comments

src/TouristInfoCards.js
```diff
@@ -0,0 +1,62 @@
...
+
+function TouristInfoCards() {
```
hanl00, line 19: I would personally stick to one way to declare react functional components, so this is more of a consistency/nitpick comment

https://github.com/CodeYourFuture/cyf-hotel-react/pull/573#discussion_r1215415298

src/TouristInfoCards.js
malkitbenning: Thank you for explaining this so clearly. I'll remember this for future.

src/TableRow.js
```diff
@@ -0,0 +1,56 @@
...
+      }}
+    >
+      <td> {props.aBooking.id} </td>
+      <td> {props.aBooking.title} </td>
+      <td> {props.aBooking.firstName} </td>
+      <td> {props.aBooking.surname} </td>
+      <td> {props.aBooking.email} </td>
+      <td> {props.aBooking.roomId} </td>
+      <td> {props.aBooking.checkInDate} </td>
+      <td> {props.aBooking.checkOutDate} </td>
```
hanl00, line 39: you can destructure the props object above to you dont need to type `props.aBooking` on every line

PR #579, opened 2023-06-11
London10 |Bedrije_Omuri|React-Hotel-app

Bedi06:


src/App.css
```diff
@@ -55,3 +63,40 @@ tr {
 .card {
   width: 18rem;
 }
+
+.rowSelected{
+  background-color: rgb(210, 207, 207) !important;
```
kitko112, line 68: !important sometimes make the job done easily but it is best not to use it. I can see the class is used by the tr, if you increase the specificity like `.table  tr.romSelected` instead of '.rowSelected', does it still work?

here is one of the reasons:
Specificity misuse: !important should be used sparingly and only in exceptional cases where there is no other viable solution. It is often seen as a quick fix to override styles, but it's better to address specificity issues by properly organizing your CSS, using more specific selectors, or re-evaluating the structure of your styles.

src/App.js
```diff
@@ -1,15 +1,48 @@
...
+      />
+      <TuristInfoCards 
+      image={manchester} 
+      header={"Manchester"}
+      text={"Home to world-famous football, renowned art galleries and a music scene that brought Oasis to centre stage – there’s lots to love about Manchester. There’s a wealth of treasures waiting to be discovered."} 
+      a={"https://visitmanchester.com"}  />
+      </div>
       <Bookings />
+      <Restaurant/>
+      <Footer props={contacts} />
```
kitko112: props means properties of the component, in your case you the property you are giving is `contacts`, so you can define your Footer component to be expecting an `contacts` array as a property, like 
`<Footer contacts={contacts} />`

here is some example of using props

```
import React from 'react';

const ChildComponent = (props) => {
  return (
    <div>
      <h2>Name: {props.name}</h2>
      <p>Age: {props.age}</p>
    </div>
  );
};
const ParentComponent = () => {
  const name = 'John Doe';
  const age = 25;

  return <ChildComponent name={name} age={age} />;
};
```

src/Bookings.js
```diff
@@ -1,18 +1,41 @@
...
-// import SearchResults from "./SearchResults.js";
-// import FakeBookings from "./data/fakeBookings.json";
+import SearchResults from "./SearchResults.js";
+
 
 const Bookings = () => {
-  const search = searchVal => {
-    console.info("TO DO!", searchVal);
-  };
+  const [bookings,setBookings]=useState([]);
```
kitko112, line 8: Nice use of useState and useEffect

src/Footer.js
```diff
@@ -0,0 +1,13 @@
+const Footer =({props})=>{
```
kitko112: define the field of the props like below is more readable, this is a followup comment of the one for App.js
```
const Footer =({contacts})=>{) 

// OR

const Footer =(props)=> {
const contacts = props.contacts;
...
} 
```

src/Heading.js
```diff
@@ -0,0 +1,11 @@
+import hotel from  "./images/hotel1.jpg"
+
+const Heading = () => {
+return <header className="App-header" style={{display:"flex"}}>
```
kitko112: the header is given a class `.App-header` already, do you think moving the `display:"flex"` into the css class is more organized?

src/Order.js
```diff
@@ -0,0 +1,20 @@
+import React ,{useState} from "react";
+import RestaurantButton from "./RestaurantButton";
+
+
+const Order = ({orderType})=>{
+
+const [orders,setOrders]=useState(0);
```
kitko112: you are storing an number, would it be better to define the variable as [order, setOrder]? giving it a plural implies it is an array.

src/RestaurantButton.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const RestaurantButton =({onClick}) =>{
```
kitko112, line 3: correct declaration of props here, nice one

Bedi06: Thank you for the feedback @kitko112

src/Order.js
Bedi06: I was thinking it will lots of orders and that's why I wrote it like that, but I understand what you mean and will make sure to implement it 💯

src/Heading.js
Bedi06: Yes, for sure, will do that.Sometimes i use this as a way to save time but i understand is not very clean n term of coding

src/App.css
Bedi06: I did it in class with a volunteer and they said the same thing about using it in very rare cases, it was more of a provisional solution, but i  will make sure to find another way of doing it. Thank you very much for taking the time for reviewing my project @kitko112

PR #582, opened 2023-06-13
london-10-shadi-fakhri-cyf-hotel-react

Shadi38:


src/Heading.js
Bedi06, line 1: you need to import the image

Shadi38: thanks Bedi

src/Order.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import {  useState} from "react";
+import RestaurantButton from "./RestaurantButton";
+
+function Order(props) {
```
llamojha, line 5: Just a small format change, try to use the same format for a function as https://github.com/CodeYourFuture/cyf-hotel-react/pull/582/files#diff-dbc58b7554ba358e418a32c54c087b909e0ea8fd1ed36aaea9229c400e228389L3 
```const Restaurant = () => {```
So it's consistent across the project.

Shadi38: thanks

PR #583, opened 2023-06-13
London10 | Seyyednavid Hejazijouybari | React | Hotel-Project(Week1 - 3)

seyyednavid:


.gitignore
BGIvanG, line 1: Removing the Gitignore (which is used to tell github what files not to include when commiting or pushing)from the project causes the node_modules to be committed which makes it hard to review and also cause the upload time to increase. This is a nice blog explaining it a bit more https://dev.to/grawl/why-you-dont-have-to-commit-nodemodules-folder-33nm.

src/App.js
```diff
@@ -1,15 +1,21 @@
 import React from "react";
-
+import Header from "./Heading"
```
BGIvanG: Its normally best practise to name the component and the file the same thing since it makes searching for them much easier.

src/Heading.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import hotelImg from "./images/hotel.png";
+
+const Heading = (props) => {
+  return (
+    <React.Fragment>
```
BGIvanG: Normally, its easier to just be able to use <> instead of React.Fragment in these situations.

src/TouristInfoCards.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import InfoCard from "./InfoCard";
+import manchesterImg from "./images/manchester.webp";
+import londonImg from "./images/london.jpeg";
+import glassgowImg from "./images/glasgow.jpeg";
+
+const TouristInfoCards = (props) => {
```
BGIvanG: If no props are being used inside of the component it doesnt need to have the props keyword here.

src/Bookings.js
```diff
@@ -1,23 +1,47 @@
...
+    setFilteredBookings(filteredBookings);
   };
 
+  const resetSearch = () => {
+    setSearchVal("");
+    setFilteredBookings(bookings);
+  };
+
+  useEffect(() => {
+    fetch("https://cyf-react.glitch.me")
```
BGIvanG: It would be good to add a .catch to this api request to handle errors.

src/Order.js
```diff
@@ -0,0 +1,17 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = (props) => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
+    setOrders((prevOrder) => prevOrder + 1);
```
BGIvanG: This is currently working great however if you wish to shorten it you can just do prevOrder + 1 without requiring the arrow function

src/SearchResults.js
```diff
@@ -0,0 +1,85 @@
...
+      setSelectedRows([...selectedRows, newID]);
+    }
+  };
+
+  const userIDHandler = (userID) => {
+    return setUserID(userID);
+  };
+
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${userID}`)
```
BGIvanG: Would be great to add error handling on this file.

src/SearchResults.js
```diff
@@ -0,0 +1,85 @@
...
+            <th>Email</th>
+            <th>Room ID</th>
+            <th>Check In Date</th>
+            <th>Check Out Date</th>
+            <th>Number of Nights</th>
+          </tr>
+        </thead>
+        <tbody>
+          {props.results.map((person) => {
+            const startDate = moment(person.checkInDate);
```
BGIvanG: For this it might be best to have the calculation above the jsx so that the logic and jsx is seprate

BGIvanG (commented):
Overall the website functions great and looks really good I liked the reset button. The code looks good aswell. Just be careful with the gitignore file since removing that caused the push of the node_modules which makes it hard to review the code in the normal way and required me to use the commits.

PR #584, opened 2023-06-13
London 10 - Paulina Wywrot - React - Week 1, 2, 3

PaulinaWywrot:


src/Heading.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+const Heading = () => {
+  return (
+    <header className="App-header">
+      <h1 className="hotel-title">CYF Hotel</h1>
+      <img src="../images/hotel.png" className="hotel-logo" alt="hotel-logo" />
```
BGIvanG: Nice work on using the images 👍 
However, the pathing seems to be slightly off. When importing images react will assume that path is from the public folder. This is a good explanation [Images in Public folder](https://dirask.com/posts/React-import-image-from-public-folder-pJAQ7D)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,29 @@
...
+        image="glasgow.jpg"
+        name="Glasgow"
+        description="Glasgow is a port city on the River Clyde in Scotland's western Lowlands. It's famed for its Victorian and art nouveau architecture, a rich legacy of the city's 18th–20th-century prosperity due to trade and shipbuilding."
+        link="https://peoplemakeglasgow.com"
+      />
+      <TouristInfoCard
+        image="manchester.jpg"
+        name="Manchester"
+        description="Manchester is a major city in the northwest of England with a rich industrial heritage. The Castlefield conservation area’s 18th-century canal system recalls the city’s days as a textile powerhouse, and visitors can trace this history at the interactive Museum of Science & Industry."
+        link="visitmanchester.com"
```
BGIvanG: Currently only the first link is working the other two arent. Visit london and manchester take the user to localhost:3000/visitmanchester.com.

src/TouristInfoCard.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const TouristInfoCard = (props) => {
+  return (
+    <div className="card">
+      <img src={`../images/${props.image}`} className="card-img-top" />
```
BGIvanG: Might need to change the path of the image here.

src/SearchResults.js
```diff
@@ -0,0 +1,27 @@
...
+  function handleIdClick(id) {
+    setCustomerId(id);
+  }
+  return (
+    <div>
+      <table class="table table-hover">
+        <TableHead />
+        <tbody>
+          {props.results.map((result) => {
+            return <TableRow result={result} handleClick={handleIdClick} />;
```
BGIvanG: Its good to add a key when creating elements from a loop.https://legacy.reactjs.org/docs/lists-and-keys.html

BGIvanG: Its good to add a key when creating elements from a loop.https://legacy.reactjs.org/docs/lists-and-keys.html

BGIvanG (commented):
This looks great! Good use of map and props. Let me know if any of the comments are unclear happy to help

src/Heading.js
PaulinaWywrot: Thank you for your comments, I've now made changes as suggested :)

PR #586, opened 2023-06-14
London class 10-Junita Lama-react module-cyf hotel-week 1/2/3

Junitalama:


src/TouristInfoCards.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+
+const TouristInfoCard = () =>{
+    return <div class="card">
```
Lauro235: In React, we can convert all instances of 'class' to 'className'. Can you think why we can't use the word 'class' as an attribute?

Lauro235 (commented):
Great work for week 1. 👍

I found it quite difficult to know what each commit was contributing to. It's helpful for reviewers to see from the commit description which portion of the exercise the commit relates to so maybe you could consider this for next week.

Well done again.

PR #589, opened 2023-06-16
London10 | Adrian Ilovan | cyf-hotel-react 

AdrianIlovan:


src/Heading.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import "./App.css";
+
+const Heading = () => {
+    return(
+        <header className="App-header">
+            <img alt="App-logo" className="App-logo" src="https://marketplace.canva.com/EAE8Xyb2xY8/1/0/1600w/canva-gold-exclusive-royal-luxury-hotel-logo-izeElzvImEY.jpg"></img>
+            <h1>CYF Hotel</h1></header>
```
moneyinthesky, line 8: It's a minor point, but just keep an eye on the formatting of your embedded html.
It would be good `</header>` was on a new line

src/SearchButton.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+
+const SearchButton = () => {
```
moneyinthesky, line 4: Nice 👍

src/TouristInfoCards.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+
```
moneyinthesky, line 2: This looks great. But it seems like there's some duplication here - the structure of each card is the same, and we are repeating that structure 3 times.
Using React, can you think of how you might be able to reduce that duplication? 😄

src/Footer.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const Footer = () => {
+
+const myProps = ["123 Fake Street, London, E1 4UD", "hello@fakehotel.com", "0123 456789"];
```
moneyinthesky, line 5: It sounds like this data should be passed into the component using React props. You can have a look at how props normally work here: https://www.w3schools.com/react/react_props.asp
Can you think of how to update the code so the data is defined in the parent component, and is passed to the Footer component as a prop?

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import FakeBookings from "./data/fakeBookings.json";
+import moment from "moment/moment";
+
+const SearchResults = (props) => {
```
moneyinthesky, line 5: This looks great!
You did a good job of passing the props in and using them in the component 👍

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
...
+                className={selectedRows.includes(booking.id) ? "selected-row" : ""}
+                onClick={() => selectedClicks(booking.id)}>
+                    <th scope="row">{booking.title}</th>
+                     <td>{booking.firstName}</td>
+                     <td>{booking.surname}</td>
+                     <td>{booking.email}</td>
+                     <td>{booking.roomId}</td>
+                     <td>{booking.checkInDate}</td>
+                     <td>{booking.checkOutDate}</td>
+                     <td>{moment(booking.checkOutDate).diff(booking.checkInDate, "days")}</td>
```
moneyinthesky, line 54: Nice 👏

src/SearchResults.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from "react";
+import FakeBookings from "./data/fakeBookings.json";
+import moment from "moment/moment";
+
+const SearchResults = (props) => {
+   const [selectedRows, setSelectedRows] = useState([]); 
+   const selectedClicks = (bookingId) => {
+    setSelectedRows((selectedClickedRows) => {
```
moneyinthesky, line 14: Awesome - great work! 😄

moneyinthesky:
Great work so far - good luck with the rest of the project 😄

src/TouristInfoCards.js
AdrianIlovan: Yes I think I can create a separate component "const Card" and pass all object info in there so I can add any other card with ease . thanks for letting me know :)

src/Footer.js
AdrianIlovan: I will make the updates , thanks :)

PR #595, opened 2023-06-18
London10 | Kristina Dubnyk | cyf-hotel-react | React

KristinaDudnyk:


src/Bookings.js
```diff
@@ -1,18 +1,40 @@
...
+      const data = await response.json();
+      setBookings(data);
+    } catch (error) {
+      console.error("The ERROR is:", error);
+    }
   };
 
+  useEffect(() => {
+    fetchData();
+  }, []);
```
kitko112: Nice use of dependencies array, empty means the logic inside the useEffect will be executed once.

src/CustomerProfile.js
```diff
@@ -0,0 +1,54 @@
+import React, { useEffect, useState } from "react";
+
+export default function CustomerProfile({ id }) {
+  // console.log("CustomerProfile props:", id);
+  const [clientData, setClientData] = useState(null);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState(false);
```
kitko112, line 7: Just a minor suggestion here, if the variable is a boolean, naming it as `isLoading, setIsLoading, isError ...` is more intuitive.

This reference is of Java but I think this convention is widely adapted 
https://geosoft.no/javastyle.html#Specific

src/CustomerProfile.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (id) {
+      fetchClientData();
+    }
+  }, [id]);
+
+  return (
```
kitko112, line 36: Instead of deciding the rendering within `return`, do you thiink moving the condition out and returning accordlingly is more readable?

```
if(loading) {
  retrun (
    <div>Loading...</div>
  )
}

if(error) {
  return (
  <div>Error</div>
  )
}

return (
  <>
    ...
  </>
 )

```

src/Footer.js
```diff
@@ -0,0 +1,17 @@
...
+  "hello@fakehotel.com",
+  "0123 456789",
+];
+
+export function Footer() {
+  return (
+    <div className="footer">
+      <ul>
+        {location.map((element, index) => (
+          <li key={index}>{element}</li>
```
kitko112, line 12: Happy to see you have the `key` attribute for each `li`

reference of key https://legacy.reactjs.org/docs/lists-and-keys.html#keys

src/Restaurant.js
```diff
@@ -1,17 +1,40 @@
...
+      Add
+    </button>
+  );
+}
+
+function Order(props) {
+  const [orders, setOrders] = useState(0);
+
+  function setOrdersFunction() {
+    setOrders(orders + 1);
```
kitko112, line 28: FYI there is another way to update the state `setOrders( (order) => order + 1)`. in this case the result is the same.

src/Restaurant.js
```diff
@@ -1,17 +1,40 @@
...
+function RestaurantButton(props) {
+  return (
+    <button onClick={props.setOrdersFunction} className="btn btn-primary">
+      Add
+    </button>
+  );
+}
+
+function Order(props) {
+  const [orders, setOrders] = useState(0);
```
kitko112, line 25: if the variable is a number type rather than an array, it is better to name it `order`, `orders` implies it is an array.

src/SearchResults.js
```diff
@@ -0,0 +1,85 @@
...
+    setSelectedProfile(id);
+  };
+
+  const handleRowClick = (rowId) => {
+    if (selectedRows.includes(rowId)) {
+      // Row is already selected, remove it from selectedRows
+      setSelectedRows(selectedRows.filter((id) => id !== rowId));
+    } else {
+      // Row is not selected, add it to selectedRows
+      setSelectedRows([...selectedRows, rowId]);
```
kitko112, line 22: Nice use of Spread `...` to update the state

PR #601, opened 2023-06-20
London10-Olha_Danylevska-React_Module-CYF_hotel_react

OlhaDanylevska:


src/App.js
```diff
@@ -1,15 +1,71 @@
...
+
+        <TouristInfoCards name="London" src="https://images.musement.com/cover/0002/49/london-jpeg_header-148518.jpeg"
+          link="https://www.visitlondon.com/" description="Welcome to London! Discover the best of London with Visit London, the official guide to England’s 
+          exciting capital. Find things to do in London, from iconic sightseeing spots and fun-filled days out to top restaurants, theatre and unmissable London events. If you’re not able to visit just yet, plan ahead to make the most of your next visit." />
+
+        <TouristInfoCards name={"Glasgow"} src="https://www.visitscotland.com/blog/wp-content/uploads/2021/10/Park_Circus_and_Kelvingrove_Park.jpg.jpg"
+          link="https://peoplemakeglasgow.com/" description="You’re guaranteed to find accommodation in Glasgow which suits your taste and budget. Whether you’re looking for something uniquely Glaswegian, a trendy hotel, a vibrant hostel or 
+          a comfortable city centre apartment, you can be assured of a warm welcome." />
+
+      </div>
```
sherif98, line 57: The TouristInfoCards component represents one card, yet it's plural (i.e. Cards not Card). I think it's better to have two components, first is TouristInfoCards which uses multiple TouristInfoCard components.

For example:

TouristInfoCards () =>
return (
TouristInfoCard(name = {"Glasgow"}
TouristInfoCard(name = {"London"}
etc.
)

You can also take the TouristInfoCards component to a new file to keep the App.js file simple and clean.

src/Bookings.js
```diff
@@ -1,21 +1,47 @@
...
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error(response.status)
+        } else {
+          return response.json()
+        }
+
+      })
+      .then((data) => {
+        console.log("data----->", data)
```
sherif98: nit: It's usually better to remove the debugging statements before you send your code for review.

src/SearchResults.js
```diff
@@ -0,0 +1,56 @@
...
+                    <th scope="col">Check Out Date</th>
+                    <th scope="col">Nights</th>
+                </tr>
+            </thead>
+            <tbody>
+                {
+
+                    allData.map(client => {
+                        let a = moment(client.checkOutDate)
+                        let b = moment(client.checkInDate)
```
sherif98: nit: it's better to have a more descriptive variable names.

let checkOutDate = moment(client.checkOutDate)
let checkInDate = moment(client.checkInDate)

sherif98 (commented):
Great work Olha!

src/App.js
```diff
@@ -1,15 +1,71 @@
...
+          data.map(line => {
+            return <li>{line}</li>
+          })
+        }
+      </ul>
+    </div>
+  )
+
+
+}
```
sherif98: What do you think of moving the Heading and Footer components in two separate files, Heading.js and Footer.js. It'll be more future proof in case the components get larger and require more work.

src/SearchResults.js
OlhaDanylevska: thanks will do)

PR #605, opened 2023-06-24
London-10_Anu Thapaliya_cyf-hotel-react

anuthapaliy:


src/Heading.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+const Heading = () => {
+   return (
+       <header>
+          <img src="https://image.flaticon.com/icons/svg/139/139899.svg" alt ="Hotel Logo" />
```
Emiliecg: There is no logo showing on the page. I think it is because this link is not a valid image. Maybe you can find a different image yourself and change the link?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+const TouristInfoCards = () => {
+   return (
+       <div className="card-deck">
+           <div className="card">
+       <img src="Glasgow image.avif" className="card-img-top" alt="Glasgow" />
```
Emiliecg: I don't see any images for the tourist info cards when I run the app locally. Where are these coming from? For example, what is `Glasgow image.avif` referring to?

src/TouristInfoCards.js
```diff
@@ -0,0 +1,34 @@
...
+           <a href="https://peoplemakeglasgow.com/" className="btn btn-primary">More Info</a>
+       </div>
+       </div>
+       <div className="card">
+       <img src="Manchester image.webp" className="card-img-top" alt="Manchester" />
+       <div className="card-body">
+       <h5 className="card-title">Manchester</h5>
+       <p className="card-text">Visit the exciting city of Football.</p>
+       <a href="https://www.visitmanchester.com/" className="btn btn-primary">More Info</a>
+       </div>
```
Emiliecg: The code for these three cards looks very similar, with some differences in the content. Have you considered using a component to avoid duplicating code?

src/App.js
```diff
@@ -1,13 +1,20 @@
...
-
+import Heading from"./Heading";
+import TouristInfoCards from "./TouristInfoCards";
 import Bookings from "./Bookings";
+import Restaurant from "./Restaurant";
+import Footer from "./Footer";
 import "./App.css";
 
 const App = () => {
+  const addresses = ["123 Fake Street, London, E1 4UD", "hello@fakehotel.com", "0123 456789"];
```
Emiliecg, line 10: This array looks like it contains data other than addresses. Can you think of a clearer name for the prop?

src/Footer.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const Footer = (props) => {
+    const { addresses } = props;
```
Emiliecg, line 4: Nice use of destructuring 😎

Emiliecg (commented):
Great start! I've added a few suggestions for improvements.
Also, could you add a Netlify link if you have one?

PR #608, opened 2023-06-26
London-10 | Iryna Lypnyk | React | cyf-hotel-react

IrynaLypnyk:


src/Bookings.js
```diff
@@ -1,18 +1,61 @@
...
+     const filteredBookings = bookings.filter(booking => {
+       const {firstName, surname} = booking;
+       const preparedSearchVal = searchVal.toLowerCase();
+       const preparedFirstName = firstName.toLowerCase();
+       const preparedSurname = surname.toLowerCase();
+       return preparedFirstName.includes(preparedSearchVal) || preparedSurname.includes(preparedSearchVal)
+     });
+     setFilteredBookings(filteredBookings);
+   } else {
+     setFilteredBookings(bookings);
```
ryankav, line 47: Nicely done, everything that follows has nothing to do with functionality as they are the exact same but want to talk about coding style. Do take these with a pinch of salt as styling varies from developer to developer, but worth sharing as it's always nice to know someone else's style.

I, personally, find that having too many variables can sometimes make the code harder to follow. So for example there are quite a few temporary variables within your filter that are reassigned. Sometimes that makes it easier to read but sometimes it can make it harder and is a balancing act to be aware of.

The second comment is that as your else branch is the case where you do nothing to the original bookings, it can be done in a different way to having two different calls to the setter. 
```js
const search = (searchVal) => {
  let searchQuery = searchVal.toLowerCase();
  let bookingResults = bookings;
  
  if (searchQuery) {
    bookingResults = bookings.filter({ firstName, surname } => {
      return firstName.toLowerCase().includes(searchQuery) || surname.toLowerCase().includes(searchQuery);
    });
  }
  
  setFilteredBookings(bookingResults)
}
```

src/Search.js
```diff
@@ -1,23 +1,38 @@
...
               <input
                 type="text"
                 id="customerName"
                 className="form-control"
                 placeholder="Customer name"
+                value={searchInput}
+                onChange={handleSearchInput}
               />
-              <button className="btn btn-primary">Search</button>
+              <SearchButton/>
```
ryankav, line 35: missing a space
```js
<SearchButton />
```

src/SearchResults.js
```diff
@@ -0,0 +1,64 @@
+import React, { useState } from "react";
+import moment from "moment";
+import CustomerProfile from "./CustomerProfile";
+
+const SearchResults = (props) => {
+  const { results } = props;
```
ryankav, line 6: can do this in place:
```js
const SearchResults = ({ results }) => {...}
```

src/Bookings.js
```diff
@@ -1,18 +1,61 @@
...
+   }
   };
 
+
   return (
     <div className="App-content">
       <div className="container">
         <Search search={search} />
```
ryankav, line 55: Can you think how you'd modify this code so that the search function lives in the search component?

Some questions that might help you think about a way you could are:
1. Is the bookings state needed anywhere else other than the search?
2. Which setState functions would you need to pass down into the search component

ryankav (commented):
Not a lot to comment on all looks good a couple of styling comments

PR #610, opened 2023-06-29
London 10 | Abubakar-Meigag | React | react-hotel-project

Abubakar-Meigag:
For review


package.json
```diff
@@ -3,6 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "axios": "^1.4.0",
+    "moment": "^2.29.4",
```
migmow, line 7: Really nice that you've started learning about `axios` 👍  
Not a good practice to push dependencies, can you do a bit of study on how to manage dependencies and exclude them from version control using Git.

src/App.js
```diff
@@ -1,15 +1,28 @@
...
   return (
     <div className="App">
-      <header className="App-header">CYF Hotel</header>
-      <Bookings />
+      <Heading />
+      <TouristInfoCards />
+      <Bookings name />
+      <Restaurant />
+      <FooterCompo
+        details={[
```
migmow: For different reasons(please research), I would suggest make App.js file as light as possible. 
What's the point of passing `details` as props to the component?
So how about if you move `details` to `FooterCompo` and then `map` on the array of details?

src/FooterCompo.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+const FooterCompo = ({ details }) => {
+  return (
+    <div>
+      <footer>
+        <ul>
+          {details.map((detail) => {
+            return <li key={detail}>{detail}</li>;
```
migmow: That's right, correctly passed `key` to each child. But, `detail` isn't always a unique key (imagine we have lots of `detail`s and it would be hard to see if each one is unique)
So how about using `{details.map((detail, index) => {
            return <li key={index}>{detail}</li>`

src/Search.js
```diff
@@ -13,11 +22,13 @@ const Search = () => {
...
+                value={searchInput}
                 id="customerName"
                 className="form-control"
+                onChange={handleSearchInput}
                 placeholder="Customer name"
               />
-              <button className="btn btn-primary">Search</button>
+              <SearchButton />
             </div>
           </form>
```
migmow, line 39: I think the search functionality is not working. You just need to define a `handleSumbit` in the parent component of `Search` component and implement `filter` functionality. Don't forget to use `prevent.default()` and then pass your `handleSubmit` to `form` tag
Would be good to do some study about [submitting form in React](https://www.w3schools.com/react/react_forms.asp#:~:text=Adding%20Forms%20in%20React&text=This%20will%20work%20as%20normal,let%20React%20control%20the%20form.)

src/TouristInfoCards.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import GlasgowImg from "./images/George Square Glasgow.jpeg";
+import LancasterCastleImg from "./images/Lancaster Castle.jpeg";
+import LondonEyeImg from "./images/London Eye.jpeg";
+
+const TouristInfoCards = () => {
```
migmow, line 6: This is all ok. But, you can make it nicer ; )
Please create an array of objects. Each object would present each city and then map on your data to display them in return.

migmow (commented):
Hi Beko - you've done a great job! you should be proud of yourself! well done 🥇 

I left some comments, let me know if you need clarify on any of them

src/FooterCompo.js
Abubakar-Meigag: done

src/Search.js
Abubakar-Meigag: done

PR #611, opened 2023-07-03
Glasgow Class 6 - Christina Mifsud - React - Hotel React

christina-mifsud:


src/SearchResultRow.js
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <tr onClick={handleClick} style={rowStyle} key={booking.id}>
+      <td>{booking.id}</td>
+      <td>{booking.title}</td>
+      <td>{booking.firstName}</td>
+      <td>{booking.surname}</td>
+      <td>{booking.email}</td>
+      <td>{booking.roomId}</td>
+      <td>{booking.checkInDate}</td>
+      <td>{booking.checkOutDate}</td>
```
hanl00, line 26: consider destructuring the booking object instead of calling it every time

PR #615, opened 2023-07-10
London10_Jan_Softa_cyf_hotel_react

softacoder:


src/App.js
```diff
@@ -1,15 +1,42 @@
...
+import Top from "./Top";
+import Heading from "./Heading";
 import Bookings from "./Bookings";
+import InfoCard from "./InfoCard";
 import "./App.css";
+import Footer from "./Footer";
+import Restaurant from "./Restaurant";
 
 const App = () => {
+  const contactDetails = [
```
sherif98, line 11: Can we move the contactDetails inside the Footer component? I don't think there is a need to define contactDetails here and pass it to the Footer component as props

src/Bookings.js
```diff
@@ -1,21 +1,125 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
+import ClipLoader from "react-spinners/ClipLoader";
 import Search from "./Search.js";
-// import SearchResults from "./SearchResults.js";
```
sherif98, line 3: Remove if not used

src/CustomerProfile.js
```diff
@@ -0,0 +1,96 @@
+import React, { useState, useEffect } from "react";
+
+const CustomerProfile = (props) => {
+  const [customerProfileData, setCustomerProfileData] = useState(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    setIsLoading(true);
+    fetch(`https://cyf-react.glitch.me/customers/${props.id}`)
```
sherif98, line 9: In Bookings.js you use async, await to do the fetch, while here you use .then() (i.e. promises chaining). I suggest using async, await here as well to make sure the code is consistent.

src/CustomerProfile.js
```diff
@@ -0,0 +1,96 @@
...
+      });
+  }, [props.id]);
+
+  if (isLoading) {
+    return <div>Loading customer profile...</div>;
+  }
+
+  if (!customerProfileData) {
+    return null; // or render a message indicating that the customer profile is not available
+  }
```
sherif98, line 32: [no action] It's nice you're separating the different return statements and not putting them all into one big return.

src/Search.js
```diff
@@ -1,23 +1,36 @@
...
               <input
                 type="text"
                 id="customerName"
                 className="form-control"
                 placeholder="Customer name"
+                value={searchInput}
+                onChange={handleSearchInput}
               />
-              <button className="btn btn-primary">Search</button>
+              <SearchButton />
```
sherif98, line 33: I feel that putting the button into a separate SearchButton component is over-engineering. The component only has a single line to declare a button and doesn't maintain any state. What do you think of just inlining the <button> here?

src/Bookings.js
```diff
@@ -1,21 +1,125 @@
...
+          </div>
+        ) : isLoading ? (
+          <div>
+            <ClipLoader color="green" size={70} />
+            <span>Loading, please wait...</span>
+          </div>
+        ) : (
+          <OutcomeSearch results={bookings} setCustomerId={setCustomerId} />
+        )}
+        {customerId && <CustomerProfile id={customerId} />}
```
sherif98, line 79: Is using the CustomerProfile component here needed? I see that you render the CustomerProfile component in the SearchResult.js line 54.

I think customerId is always undefined and thus it never renders the component here. Please take a look and let me know what are your thoughts on this. Thanks

sherif98 (commented):
Overall looks great Jan! I added a couple of comments.

PR #617, opened 2023-07-11
LONDON_10 || SAIM KORKMAZ || HOTEL REACT APP

nsaimk:


src/App.css
```diff
@@ -55,3 +55,66 @@ tr {
...
+}
+
+.new-booking{
+  border: 0.5px solid rgb(184, 183, 183);
+  padding: 25px;
+}
+.new-booking button{
+  background-color: rgb(215, 213, 74);
+}
+.highlighted0{
```
kitko112, line 76: What elements are using these classes? if the styles are the same, one css class is enough.

if you want to style particular item in a list of item, you can try below class
li:nth-child(odd)  // item with odd index
li:nth-child(even)  // item with even indexx
li:nth-child(2) // All [<li>](https://www.dofactory.com/html/li) elements that are second child of their parent

https://www.dofactory.com/css/ref/selectors

src/Bookings.js
```diff
@@ -1,18 +1,113 @@
...
+        }
+        return res.json();
+      })
+      .then((data) => {
+        setBookings(data);
+      })
+      .catch((error) => {
+        setError(error.message);
+      });
+  }, []);
```
kitko112, line 33: Nice use of dependencies array in useEffect

src/Bookings.js
```diff
@@ -1,18 +1,113 @@
...
+        return (
+          row.firstName.includes(searchVal) || row.surname.includes(searchVal)
+        );
+      })
+    );
+  };
+
+
+  const handleChange = (e) => {
+    setNewBooking({ ...newBooking, [e.target.name]: e.target.value });
```
kitko112, line 48: A short and clear way to update the booking, Nice one

src/components/CustomerProfile.jsx
```diff
@@ -0,0 +1,26 @@
...
+  const [customerInfo, setCustomerInfo] = useState("");
+  useEffect(() => {
+    fetch(`https://cyf-react.glitch.me/customers/${id}`)
+      .then((res) => {
+        return res.json();
+      })
+      .then((data) => {
+        setCustomerInfo(data);
+      });
+  }, [id]);
```
kitko112, line 14: Nice one, only the id change will trigger the useEffect

src/components/Order.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { useState } from "react";
+import RestaurantButton from "./RestaurantButton";
+
+const Order = ({ orderType }) => {
+  const [orders, setOrders] = useState(0);
+  const orderOne = () => {
```
kitko112, line 6: FYI

another way to use the setOrders
```
setOrder((order) => order + 1)
```

PR #618, opened 2023-07-13
BAKI TUNCER/LONDON-10/CFY HOTEL

batuncer:


src/App.js
```diff
@@ -1,15 +1,27 @@
...
-
 import Bookings from "./Bookings";
 import "./App.css";
+import Heading from "./components/Heading";
+import TouristInfoCards from "./components/TouristInfoCards";
+import Footer from "./components/Footer";
+import Restaurant from "./components/Restaurant";
 
 const App = () => {
+  const addresses = [
```
BGIvanG, line 10: This variable makes it seem like the list contains only addresses. However it has an email and a phone number. It might be better to rename it to something like details.

src/Bookings.js
```diff
@@ -1,19 +1,52 @@
...
-  const search = searchVal => {
-    console.info("TO DO!", searchVal);
+  const [bookings, setBookings] = useState([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState(null);
+
+  useEffect(() => {
+    console.log("Page loaded");
+    setIsLoading(true);
+    setError(null);
```
BGIvanG, line 13: Really nice use of error handling here on page load. However, you shouldnt need this since you already have the default set to null.

src/components/SearchResults.js
```diff
@@ -0,0 +1,100 @@
...
+        </tbody>
+      </table>
+      {selectedCustomerId && (
+        <CustomerProfile customerId={selectedCustomerId} />
+      )}
+    </div>
+  );
+};
+
+const CustomerProfile = ({ customerId }) => {
```
BGIvanG, line 54: You used the Id really well here to allow for searching. Only suggestion would be to move this into a separate file so it becomes its own reusable component if it needs to be used elsewhere. Similar to the other components in the app.

src/components/Restaurant.js
```diff
@@ -0,0 +1,38 @@
+import React, { useState } from "react";
+
+const Order = ({ orderType }) => {
```
BGIvanG, line 3: This is great separation making them all into smaller components however it still might be best to split them into separate files to keep them smaller and easier to read.

src/components/SearchResults.js
```diff
@@ -0,0 +1,100 @@
...
+      <table className="table table-striped">
+        <thead>
+          <tr>
+            <th>Title</th>
+            <th>First Name</th>
+            <th>Surname</th>
+            <th>Email</th>
+            <th>Room ID</th>
+            <th>Check-in Date</th>
+            <th>Check-out Date</th>
```
BGIvanG, line 21: For task 6 to add the difference of nights You installed moment library which is great. The docs provide some guidance on calculating the difference here https://momentjs.com/docs/#/displaying/difference/

src/components/SearchResults.js
```diff
@@ -0,0 +1,100 @@
...
+            <th>Actions</th>
+          </tr>
+        </thead>
+        <tbody>
+          {bookings.map((booking) => (
+            <tr key={booking.id}>
```
BGIvanG, line 27: The map works perfectly here and it seems to function as expected. It just seems to be missing lesson 2 task 15. You can achieve it by adding a onClick here and a function to change its class

src/components/CustomerProfile.js
```diff
@@ -0,0 +1,33 @@
...
+
+const CustomerProfile = ({ customerId }) => {
+  const [customerProfile, setCustomerProfile] = useState(null);
+
+  useEffect(() => {
+    if (customerId) {
+      fetch(`https://cyf-react.glitch.me/customers/${customerId}`)
+        .then((response) => response.json())
+        .then((data) => {
+          setCustomerProfile(data);
```
BGIvanG, line 11: Would be good to add a catch here aswell and some error handling similar to the other api requests. Then you can use that to display an error message.

BGIvanG (commented):
Great Work on working with the Apis and handling the error. Just a few comments mostly about moving components out into their own file. its normally best as it keeps the project clean and makes the components more reusable.

